PR #26, opened 2021-11-19
Major refactoring

SukuWc:
## Tested
 - Profile save
 - Profile load
 - Element copy
 - Element overwrite
 - Basic config blocks add copy drag


kkerti:
Very nice work!

PR #240, opened 2023-05-04
Fixed crash on unknown midi command

elsoazemelet:


SukuWc:
@elsoazemelet  code looks good! 
@narayb please test MIDI sync messages in the MIDI monitor panel!

PR #247, opened 2023-05-12
Fixed clipping and added text truncation for Action blocks

elsoazemelet:
On smaller displays, some action blocks input field title is to long to be rendered in one line, causing the misalignment of the layout of these action blocks.

This fix contains the truncation of these titles, meaning when a title is too long, the title will be rendered in the following pattern: 
```
"Some Long Input Field Title" ⟶ "Some Long In...". 
```

The layout is intact on smaller displays.

A new issue was found during testing, this PR contains the fix for that too.
The issue is that the layout of some action blocks was coded poorly, and the automatic layout resizing had a implicitly defined minimum size. When the width of the Config menu was smaller than this minimum width of the action block, the action block "broke out" of the Config menu. See the following image:

image

The layout of MIDI MIDI14 and KeyboardMacro action blocks were reworked.

### DIFF
--- a/src/renderer/config-blocks/LedColor.svelte
+++ b/src/renderer/config-blocks/LedColor.svelte
@@ -514,7 +514,9 @@ A -> B : AB-First step
   <div class="w-full flex">
     {#each [scriptSegments[2], scriptSegments[3], scriptSegments[4]] as script, i}
       <div class={"w-1/3 atomicInput "}>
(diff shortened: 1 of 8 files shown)


narayb:
The tested Action Blocks now all truncate text properly and it doesn't affect the placement of text boxes anymore.

PR #248, opened 2023-05-15
Fixed Grid <=> Editor config synchronization and reactivity for removing modules

elsoazemelet:
This PR contains the following changes:
- CodeBlocks can not be committed if the char limit is reached
- Action blocks can not be pasted if the to be char limit would be reached
- Action blocks can not be added  if the to be char limit would be reached
- Active changes will be reduced with the amount of changes on the removed module (e.g.: 2 modules, 1-1 changes on both, one gets removed => 1 active changes instead of 2 is displayed)
- selectedElement is set to default if the current selectedElement was on the removed module


narayb:
Changes tested work as they are supposed to.

PR #255, opened 2023-05-22
Store button is always displayed

elsoazemelet:
- [x] Check if Store and Discard buttons are always displayed
- [x] Store and Discard buttons are disabled when 0 active changes are present
- [x] Check reactivity when the active changes are on a second module, and then disconnect  the module. Reaction should be that all changes are lost, so the Discard and Store buttons should be disabled.


narayb:
Tested on win10. Working as intended. @elsoazemelet

PR #263, opened 2023-06-14
Configuration management is reworked from the ground-up

elsoazemelet:
**Goals of this task**

- Unifying the representation of a config list and the action blocks in code
- Creating a common interface through which configuration scripts can be sent to Grid
- Simplifying algorithms used during configuration manipulations
- Remove unnecessary code, and simplify code
- Remove unnecessary complicated ways of rendering an Action Block (in code)

Keep in mind that all of these changes were purely under the roof type of changes, so the QAs purpose is to assure that all the vital functionalities of the editor is intact.

**Possibly affected functionalities**

- [x] Copy
- [x] Paste
- [x] Cut
- [x] Merge to Code
- [x] Remove
- [x] Select All
- [x] Single Selection
- [x] Multiple Selection
- [x] Loading configs
- [x] String configs
- [x] Clearing configs
- [x] Switching events
- [x] Switching UI/System events
- [x] Copy All
- [x] Paste All
- [x] Runtime error displaying when LUA is not OK
- [x] Syntax checking
- [x] Input field validation
- [x] Drag and drop
- [x] Drag and drop ESPECIALLY edge-cases
- [x] Normal adding of configuration
- [x] Insertion type adding of configuration
- [x] Prompt type adding of configuration
- [x] Reactivity of config script in the Debug Monitor
- [x] Length checking when modifying a config (the displayed length in debug monitor and actual length should be synchronous [400 chars])
- [x] Suggestions are still intact (including local definitions)
- [x] Last opened action blocks is open when changing an event
- [x] Drag and drops animation is intact

**NOTE:** This is not all the affected areas, but most of it, it should give a good understanding of the magnitude of this PR, and the areas which are affected.


narayb:
Each feature not mentioned below and has a checked mark in the checklist above, has been tested and works as expected.

- [x] **copy, paste, cut, delete, merge to code, select all** action runs but only shows the change after refreshing the editor with ctrl+r (refresh render function missing?)
- [x] trying to add Actions works, but similarly to the issue above only gets displayed after a forced refresh
- [x] trying to open MIDI monitor sidebar causes app to crash

image

- [x] **active changes** descriptor number continually increases when interacting with changed control element
- [x]  these control element changes also cause the debug monitor to display the current config as a new config each time a CE interaction happens
- [x] trying to switch Pages while having an unstored configuration doesn't display an error message as it did previously
- [x] trying to switch the MIDI Action to 14bit or SysEX throws an error and does not work however adding the Action Block is possible but still crashes the app

image

- [x] trying to open the MIDI block crashes the app 

image

- [x] trying to open the code editor or the code block crashes the app

image

- [x] hitting commit on all configs in code editor results in the following

image
image


- [x] sometimes control elements will not render on screen until receiving manual input from outside while throwing this error message

image
image

narayb:
Each feature not mentioned below and has a checked mark in the checklist above, has been tested and works as expected.

- [x] active changes descriptor number continually increases when interacting with changed control element
- [x] these control element changes also cause the debug monitor to display the current config as a new config each time a CE interaction happens

PR #270, opened 2023-06-23
Fixed tool tip transparency for BtnAndPopUp

elsoazemelet:
By hovering over the actions when they are disabled (cut, copy, etc.) the tool tip now should function as expected, without transparency.

Bug was:
image


narayb (commented):
Changes ok. Bug fixed, feature performs as it should now.

PR #272, opened 2023-07-04
Dynamical character count is added to CodeBlock code editor

elsoazemelet:
- [x] Char count increases with each key stroke
- [x] Char count in the debug menu does not change until editor is closed
- [x] Code can be committed and stored
- [x] A modified code can be undone by exiting the code editor
- [x] Only syntax correct code can be committed
- [x] Code over 400+ chars can not be committed
- [x] After committing a code the debug window shows the same char count
- [x] Syntax error codes char count can not be determined and is marked with a "?"
- [x] The code is indeed minified and this shows in the char count


narayb:
All features not mentioned below have been tested and found OK.

In case of going over the character limit, the code editor shows "?/400" as if the code was syntax-incorrect. Expected behavior would be showing how much over the limit the charactercount is.

image

elsoazemelet:
Fixed PR issues

narayb:
Previous issues fixed.

- [x] In case of opening the code editor the first time, an active change is generated even when the editor is closed without changes. 

- [x] During the above event, switching to a different control element on the UI and then trying to switch back to the one with teh code block, causes the app to lose connection to Grid and throw the following error message:
Pasted image 20230705130208
This leads to the following line:
Pasted image 20230705130242

elsoazemelet:
Fixed PR issues

narayb:
All crashes are fixed when switching between Code Blocks with active changes.

PR #275, opened 2023-07-05
Added 3-way-switch to change tracking of grid module interactions

elsoazemelet:
Removed the "Physical interaction tracking" checkbox feature from under Preferences.
Now it can be found as a 3-way-switch in the grid layout (middle panel) in the editor.

The options of this switch are the following, and these should be tested:

- [x]  OFF: By interacting with the grid modules, no element and no event should be selected automatically.
- [x] Element: By interacting with the grid modules, only the element should change, but not the event.
- [x] Event: By interacting with the grid modules, everything should work as normal, the currently selected element should be the element lastly interacted with, and the event should be the proper event (e.g.: on fader slide it should be the potmeter event).

Additional features that should be tested:

- [x] When on Element tracking, the selected event was e.g.: encoder event, when switching on a fader element, the selected element should be INIT. The reason is that when the selected event is non existent on a newly interacted element, then the selected event always should be INIT.
- [x] What happens on grid module removal.


narayb:
- [ ] [UX] when switching between different element types in the 'Element mode' of Interaction Tracking, the tracking keeps defaulting to init which is still counterintuitive, expected should be the similar type of interactable element (e.g.: switching from encoder button to a potmeter defaults to init event, but would be more logical for it default to the potmeter event)
~~- [ ] in ALL modes when switching between to different element type by clicking on the UI of Editor, incompatible changes result in errors and the incompatible event type causing the Editor ui to not render anything under that element~~ MOVED TO Backlog

PR #276, opened 2023-07-06
Fixed unknown event error (example: when changing from Encoder to Fader)

elsoazemelet:


narayb:
Code Block render breaking  is fixed. Tested on Mac.

PR #293, opened 2023-07-11
Added the option to enable/disable analytics

elsoazemelet:
- [x] In Preferences, analytics gathering can be enabled/disabled
- [x] On the Welcome Screen, the same option can be seen
- [x] Analytics gathering is disabled until first accepted on the Welcome Screen
- [x] On the welcome screen and in the preferences, the analytics gathering policy can be opened

TODO:
The analytics gathering policy webpage is not yet done


narayb:
Can't test if it should be disabled on first startup. Old install was on for me.

PR #294, opened 2023-07-12
Fixed MIDI block replacing and visual responsivity

elsoazemelet:
- [x] MIDI blocks can be changed (MIDI to MIDI14, MIDI to MIDISysEx, etc.), and the work properly
- [x] When resizing the window to extremely narrow, the visuals remain intact, especially to the MIDI type tabs (on top of the block, the trapezoid buttons)


narayb:
Both fixes are confirmed working on win10.

PR #296, opened 2023-07-13
Background running is now optional in preferences

elsoazemelet:
- [x] Now by default by closing the app window, the whole application is closed (no need to be closed from the tray)
- [x] This behaviour (the old one) can be turned on/off in the Preferences


narayb:
Fully works as intended on win10.

PR #297, opened 2023-07-13
Added fading in/out for Tracker when logger messages are displayed

elsoazemelet:
- [x] When receiving a Logger message (on the bottom center of the screen, for example when trying to change pages during active changes), the visual display of these messages do not clash with the Tracking panel (OFF, Elements, Events bar on the bottom).
- [x] Works with multiple messages too
- [x] It should be displayed again, when all logger messages disappear
- [x] Has a cool animation, but it should be approved by QA


narayb:
UX - animation takes a bit too long, maybe clicking on these Logger messages could dismiss them?

QA - works, looks nice!

PR #298, opened 2023-07-13
Fixed bug of changing to profile cloud when selecting preferences

elsoazemelet:
- [ ] When selecting a different tab than Profile Cloud, and then switching to Preferences tab, the selection of the Preferences tab does not revert the selected left side tab to Profile Cloud


narayb:
Yup, fixed.

PR #299, opened 2023-07-13
Added error message when during active changes the utility button is pressed

elsoazemelet:
- [x] Now when switching pages during active changes with the utility button, the same error message is displayed as when the page changing is done on the UI (by the page selector buttons)


narayb:
Works as described.

PR #310, opened 2023-07-20
SUKU module architecture displayed on device

SukuWc:


SukuWc:
@narayb please check if visual design is acceptable!

kkerti:
Visual design is correct, tested with multiple modules in different orientations.

image

PR #317, opened 2023-08-01
Fixed double change and added feature of change only occurs on focus change

elsoazemelet:
- [x] Active changes now only increment happen, when an input field on an action block has changes, and looses focus.
- [ ] This functionality should work on all blocks including input fields.

### DIFF
--- a/src/renderer/config-blocks/Comment.svelte
+++ b/src/renderer/config-blocks/Comment.svelte
@@ -20,9 +20,9 @@
 </script>
 
 <script>
-  import { onMount, createEventDispatcher, onDestroy } from "svelte";
+  import { createEventDispatcher, onDestroy } from "svelte";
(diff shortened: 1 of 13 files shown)


kkerti:
Does not work with following action blocks (changes are triggered continously)
- [x] Encoder Mode
- [x] Button Mode
- [x] Potmeter Mode
- [x] IF, ELSE IF block
- [x] Comment block
- [x] Element name
- [x] Timer Stop
- [x] Lookup

kkerti:
Please review:

- [x] if
- [x] elif
- [x] lookup
- [x] comment block - doesn't trigger active changes
- [x] element name - doesn't trigger active changes
- [x] keyboard

kkerti:
@elsoazemelet please run formatter!

kkerti:
Ok!

PR #318, opened 2023-08-01
Reworked/improved feedback sending window, and added analytics

elsoazemelet:
Feedback modal window is reworked. Now it should be reactive for window size changes, and should never break the layout. New animation is added when feedback is sent.

- [ ] Feedback is sent to discord.
- [x] Feedback appears in analytics too.
- [x] Window closes after sending feedback in a few seconds
- [x] After sending a feedback, the same feedback in the same window can not be sent multiple times.


kkerti:
Please add correct webhook URL, then good to go.

PR #320, opened 2023-08-02
Fixed Keyboard Macro Action blocks error displaying

elsoazemelet:
Keyboard macros only take effect after storing. When a keyboard macro is added or an existing is modified, a warning message is displayed (Keyboard macros will only take effect after storing), and can not be used until storing. When stored the macros should work and the warning message is dismissed.

- [x] Warning is displayed when a stored keyboard is modified (or a new is added)
- [x] When stored, this warning is dismissed

### DIFF
--- a/src/renderer/config-blocks/Macro.svelte
+++ b/src/renderer/config-blocks/Macro.svelte
@@ -67,6 +67,15 @@
   let loaded = false;
   let macroInputField;
 
+  let isStored = true;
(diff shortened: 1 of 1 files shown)


kkerti:
Ok!

kkerti:
Second thought: 
- please check coloring from "error" to "warning" (use orange-ish color)
- please check what happens when multiple keyboard actions are present

Issue #322, opened 2023-08-02
Fix always run in the background crash on macOS


kkerti:
The `always run in the background` when ticked off resulted in crash under MacOS.

PR #323, opened 2023-08-02
Always run in background exit crashed on macOS 

kkerti:
#322 
When always run in the background was ticked OFF, closing the app resulted in crash. Fixed by adding logic to "window-all-closed" under darwin.

App should stay on tray under linux/windows and stay on dock on macos, when running in the background is active. When inactive, quit the app for good. Default behaviour should be on?

- [x] Test on Windows
- [x] Test on Linux
- [x] Test on MacOS

### DIFF
--- a/src/electron/main.ts
+++ b/src/electron/main.ts
@@ -226,7 +226,6 @@ function createWindow() {
       mainWindow = null;
     } else {
       evt.preventDefault();
-
       // only hide, keep in the background
       const keepRunning = store.get("alwaysRunInTheBackground");
       if (keepRunning === true) {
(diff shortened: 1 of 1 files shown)


src/electron/main.ts
```diff
@@ -600,4 +606,7 @@ app.on("activate", () => {
 });
 
 // termination of application, closing the windows, used for macOS hide flag
-app.on("before-quit", () => (app.quitting = true));
+app.on("before-quit", () => {
+  app.quitting = true;
+  console.log("before quit");
```
danim1130, line 611: Looks sketch

danim1130 (approved):

PR #350, opened 2023-08-10
Refactored out utils.js

### LINKED ISSUE #283: Refactor out and delete utils.js
- Delete unnecessary functions (isJson, etc.)
- If possible declare the remaining content of utils.js in its place of use.
- If it is not possible (due to multiple usage), then define a new JS file and put it next to the most logical use of it.

elsoazemelet:
Affected blocks:

- LED Animation Start
- LED Animation Stop
- LED Color
- LED Intensity
- Lookup
- MIDI/MIDI14/MIDISysEx
- Mouse Button
- Mouse Move
- Timer Start

Check that these blocks parameters still can be edited, the stored and reloaded correctly.

Closes #283


narayb:
Hard to tell if it really is ok, but no errors were found during testing.

PR #353, opened 2023-08-14
SUKU portstate visualization debugging feature added

SukuWc:


SukuWc:
- under developer setting in preferences enable port state visualization!
- when multiple modules are connected a green box shows active connection
 
image

narayb:
Works nice on new FW on both ES3 and D51

image


In this case there could be some kind bug, since the module reported that it was connected but now it clearly isn't but is still displayed with it's interfaces green.

PR #354, opened 2023-08-14
Bundle profile cloud on editor release, allow offline functionality

### LINKED ISSUE #327: [BLOCKER] Cached version of profile cloud buldned with editor

danim1130:
When creating a new release, Github Actions will checkout the current profile-cloud repository and build the latest tagged commit. The resulting project will be bundled with the Editor.

If loading the online version of the profile cloud fails, the Editor is able to use the bundled version instead.

- [ ] When no internet connection is available at first start, 'Load Offline' button should be visible in the profile cloud pane
- [ ] Pressing the button will load the bundled profile-cloud version


narayb:
I might be misunderstanding this feature, but on the profile cloud sidebar, even when starting in offline there is only a login prompt, nothing else. If offline, I'm unable to log in of course.

narayb:
When loading the offline mode, clicking the module after makes the profile cloud sidebar disappear again.
image

danim1130:
Issue fixed!

PR #355, opened 2023-08-14
SUKU timer source action block added

SukuWc:
This allows grid to use midi clock from DAW as timer frequency referece clock.
- 0: Default means internal clock (milliseconds)
- 1: Midi beat clock

image

- This setting works on a per element basis, so different elements can have different clock sources!
- Test with multiple modules

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "grid-editor",
-  "version": "1.2.39",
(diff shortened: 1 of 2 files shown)


SukuWc:
Use latest nightly firmware to test!

narayb:
I get it now. Works as it should.

When releasing new version of Editor, new feature requires this FW release as well!

Issue #345, opened 2023-08-08
Move developer settings (and other advanced features) under an Advanced field


elsoazemelet:
Move option field from preferences under an advanced field.

- Advanced field and the options can be enabled with a checkbox
- The checkbox and its label should be located at the bottom of preferences
- Advanced preferences can be toggled
- Make the design fit the current layout seamlessly

PR #364, opened 2023-08-22
#345 Move developer settings (and other advanced features) under an Advanced field

kkerti:


danim1130:
- Radio group background color?
- Port state / web socket developer options / etc?

SukuWc:
Also please restore profile cloud options

kkerti:
Added missing options, known issue: centering of the radio and checkboxes.

PR #383, opened 2023-09-04
Fixed indentation calculation for Action Blocks and  fixed block naming and display

elsoazemelet:
The indentation is now fixed for double or more sided action blocsk (e.g.: If Else, If ElseIf Else, Button up/down, etc.). All the missing wrong block icons/titles are now fixed.

Check all these button/encoder/if blocks, the indentation correctness, and check all the action blocks in the picker menu and in the config list that it renders with correct titles and icons.


narayb:
Looks ok, Else if  should gain the same "Elif" naming scheme in the Action Block picker window.

PR #409, opened 2023-09-13
Added auto scroll for configuration list when dragging configs

elsoazemelet:
When dragging an action block in a configuration list that is scrollable (the total height of the opened/unopened action blocks are greater than the available space for the config list), when the cursor is at the top or bottom of the available space of the config list, the config list auto scrolls towards the top/bottom.

- Test for the feeling of the dragging with an opened action block
- and an unopened action block


SukuWc (approved):
Nice and clean

narayb:
If the opened block is bigger than the list itself, the feeling is not great. I would suggest automatically collapsing the blocks visually when trying to drag an opened block.

The feeling is OK. 

When dragging with the mouse ouside of the list window the dragging is stopped, this feels bad.

PR #413, opened 2023-09-14
Fixed orientation and selection for profile and preset overlay

elsoazemelet:
When a module is connected it is rotated by the:

1. orientation of the connected module (by connecting the module in a 90/180/270 degree relative to the main module)
2. global orientation that is set in the preferences

These two rotation types are combined for the calculation of the main view in the editor.

The bug was that the preset/profile load overlay was rotated with the modules.

Test for:

- Different module connection rotations
- Different global rotation in preferences
- With combinations

In every combination the preset/profile overlay should be consistent and facing upwards (readable with text).


SukuWc (approved):
Ok

narayb:
Works OK.

PR #415, opened 2023-09-14
Toggle elements now opened after event changes

elsoazemelet:
Now not only the lastly toggled action block is auto-toggled when changing an event, but all the blocks that were opened on the previous event, if possible.

Mind that, when on event #1 a (for example) code block was opened, changing first to event #2 (where no code blocks are present) nothing will open. But when changing to event #3 (where a code block is present) the code block is toggled, even though on event #2 no code blocks were interacted with.

This behaviour can lead to confusing situations, but the solution is consistent with the requirement of all toggled action blocks should be auto-toggled when changing events.

Further discussions of functional behaviour could be necessary, if QA or PO initiates it.

### DIFF
--- a/src/renderer/main/panels/configuration/Configuration.svelte
+++ b/src/renderer/main/panels/configuration/Configuration.svelte
@@ -47,7 +47,7 @@
   import { selectedControllerIndexStore } from "/runtime/preset-helper.store";
 
   const configs = writable([]);
-  let lastOpenedElementsType = undefined;
+  let lastOpenedElementsType = [];
   let events = { options: ["", "", ""], selected: "" };
   let elements = { options: [], selected: "" };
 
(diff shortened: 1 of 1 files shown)


SukuWc (approved):
Ok, clean!

narayb:
## Functionality issue, confusing
This looks OK at a glance, but when creating a new action block on a fully opened event, all Action Blocks get collapsed. Switching to an element where a block like this is toggled open and then switching back causes all blocks to be open properly.

They should stay opened (if they were open before adding a new block) and the new block should arrive closed.

## UX issue
Only additional concern from UX pov is that in smaller resolutions and window sizes, the newest and usually the block the user is working on, appears at the bottom of the list by default. If the blocks are all open, on an event switch the bottommost action might require scrolling, even though that's what the user is working on.

PR #417, opened 2023-09-15
SUKU element name overlay rotation fixed, TEK2 overlay added

SukuWc:
element name rotation fixed, 
code cleanup + 
TEK2 element overlay implementation


narayb:
Elename confirmed fixed. TEK2 not tested.

Issue #431, opened 2023-09-25
Loop special eaction block (for TEK2 led handling)


elsoazemelet:
Make wonders, create a loop block and its flavours:

- FIRST: for loop
- SECOND: think about while and do while

Issue #427, opened 2023-09-23
Special action still does not show text "Rotate Left"


kkerti:
<img width="489" alt="image" src="https://github.com/intechstudio/grid-editor/assets/47832952/3f05586e-eb26-4f96-acd4-21c6ea4cd397">

Issue #426, opened 2023-09-23
Incorrect display of encoder action blocks


kkerti:
<img width="1479" alt="image" src="https://github.com/intechstudio/grid-editor/assets/47832952/1533e08c-c26a-4364-97b8-e335d847046b">

elsoazemelet:
Old error:
Special right / left encoder rotation action with an if block added triggers error

Solution:
No repro, hovewer fix desgin!!!

PR #446, opened 2023-10-02
[MEGA]: Added For Loop action block for editor and reworked ConfigManager

elsoazemelet:
For Loop is added as a new action block. The implementation caused the following changes:

- Added new action block for action block picker
- Reworked action block toggle functionality
- Now every type of action block has it's own "face" (untoggled state). These types are: 
   - Regular action block face (single, "normal" action blocks)
   - If type action blocks with input field on its face
   - For Loop face
- Added control properties for individually definable behaviour for action blocks (can it be toggled, moved, selected, etc.)
- Reworked batch selection for composite (if and loop type blocks with action blocks in between)
- Reworked indentation

These changes  require the following tests, by individually testing each action blocks:

- [x]  Dragging, the face can be dragged almost everywhere
- [x] They can easily be opened
- [x] Loop is understandable how can it be toggled
- [x] Every input field can be edited
- [ ] All the icons and names, etc. show properly
- [ ] All the indentation is rendered correctly
- [ ] The For loop works properly
- [ ] For loop can be edited from the toggled/untoggled state
- [ ] Syntax checking is still intact
- [ ] Selection is still intact
- [ ] For loop can be stored
- [ ] For loop backwards compatibility, what happens if it is used with older versions of editor

Closes, #431 , #426, #427


narayb:
- [x]  when closing-opening the loop blocks, the whole block resizes and the button moves (Persisting)
- [x]  loop block lacks icon (Discuss with if)
- [x]  loop is hard to drag when opened or is not able to be dragged at all 
- [x]  if block is not draggable at all
- [x]  when dragging cascaded blocks an indicator appears to put the group of blocks inside itself, which it can't do  ! (Persisting) 
[Pasted image 20231003142052](https://github.com/intechstudio/grid-editor/assets/74784772/61f57824-a6c9-466d-966f-375df2bb3d99)
- [x]  use a switch for the block if possible Pasted image 20231003142341  its something we used before
- [x]  loop changes size when opened: 
 Pasted image 20231003142448  and this is after opening: the indentation breaks for the top 
  Pasted image 20231003142525
- [x] bad syntax highlight is barely visible in front of an orange background  Pasted image 20231004120818
- [x] if the opened view is empty and then the block is closed, bad syntax becomes store-able AND the block shows Infinite, but still works as if the last correct syntax was working and doesn't repeat infinitely Pasted image 20231004121048
- [x] bad syntax still runs as if it was correct
- [ ] trying to load a config IN OLD VERSION of Editor with for loop block results in editor freeze Pasted image 20231004121504

PR #450, opened 2023-10-12
Reworked tabbing between input fields to support active change detection

### LINKED ISSUE #412: Tabbing between input fields on action blocks will only send the first change to Grid
Reproduction: 
- open midi block,
- click channel, enter new value like "8"
- tab
- enter new value
- tab
- enter new value
- click outside of action block

The result is a single active change. Seems like the action block resets it's state back, so Editor stays in sync, but user may get frustrated. Probably it will be solved by using a component library.

elsoazemelet:
Closes #412


narayb:
nice!

PR #454, opened 2023-10-16
[MEGA] Added RAW data / code refactor / Drag-zone threshold

### LINKED ISSUE #456: Refactor information override out
New encoder PRLR with different shorts
Follow up: RAW code block

elsoazemelet:
Closes #456 #457 #469 #458

**Review:**
Review the new added RAW action block that can be triggered in this PR build as adding a no longer supported action block, that is on the encoder event, the Push&Rotate LR action block.

Follow the steps that it offers, and review its design and feel.

**Test:**

- [x] The threshold of the dropzones which is now added, discuss the feel of it
- [x] The new Push&Rotate LR action block, that it is intact and works as intended (after replacing the RAW blocks, and before replacing them)
- [x] The MIDI replace when clicking on tab works as intended, and it can be stored. Same with the replaced RAW blocks.
- [x] The non-selectable composite special blocks parts are now indicated with a red checkbox. Check all these special action blocks and its parts that they are indeed correct, and if they can be moved. (and if possible is that correct? Double check all these stuff..) 
- [x] Converting RAW action block to CodeBlock
- [x] Merge to code is not broken
- [x] CodeBlock edit is not broken


narayb:
Pushrot block. Check visual thing! Is it good?

narayb:
- [ ] cursor scrolls the Action Chain even when not dragging anything, or when mouse cursor is outside of application window
 - [x] not very intuitive what actually happened, like the order of the raw blocks representing the block conditionals
 - [x] parts of a conditional block being code and parts of it being blocks is even more confusing, especially when now code based conditionals are edited and then the names of the block like "just rotate right" become completely false, kind of breaking the purpose-made nature of blocks
 - [x] code block is kind of broken, when opening a block from converting the code will be short code, after committing it, it will become an empty block

PR #473, opened 2023-10-27
Regular MIDI action block has a prototype closed face

### LINKED ISSUE #465: MIDI block preview on its face

elsoazemelet:
Insert a MIDI block (regular) and review it's new design when closed.

Closes #465


narayb:
image
- [ ] I feel this either could be even more minimal or explain more, but it falls in the middle of not saying enough, but does that with a lot of characters
- [ ] it should display midi channel and value both if trying to be informative
- [ ] or display names instead of values when trying to be readable (CC instead of 176, Note instead of 144 etc )
- [ ] or should only display values if trying to be minimal (as a midimonitor `MIDI TX: 0 176 12 val` )
- [ ] values instead of name in the same typeface when compared to other blocks looks out of place

PR #498, opened 2023-11-17
Removed new lines from LUA arrays when formatting

### LINKED ISSUE #476: Do not break each value of an array into a new line please!
Each value should be in the same line for readability in the code editor.

elsoazemelet:
Closes #476


narayb:
Hands down best update.

PR #502, opened 2023-11-27
SUKU system element shape cutout

SukuWc:
This PR contains the following features:

- Reworked system element selection, now it is accessible from clicking on the firmware version area.
- Reworked active change design for control elements and system element.
- Fixed overlay display for user input, and preset/profile loading

Test:

- [x] Profile loading
- [x] Preset loading
- [x] Preset/Profile and Control Name overlay for rotation and scale (ignore CSS scaling glitch?)
- [x] Look of active change design
- [x] Look and feel of system element selection design
- [x] User Input on the module when Preset/Profile overlay is shown
- [x] Since now user input is not triggered for preset/profile loading, test the feel of it. Is it confusing? (especially when loading a preset multiple times)
- [x] Try to break the overlay displays by getting into an unexpected state, by doing multiple actions (e.g.: something disappears when it should not, or something is stuck when it should not)


narayb:
## Testing Overlays


- [x] Module orientation is not correctly recognized and displayed. The EF44 on that pciture is connected upside down, but appears upright. Loading a preset in this misrepresented state. The preset load overlay appears in the correct orientation of the module, but because the module is not rendering upside down, the text is upside down as well?

kép



- [x] Sometimes parts of the Selection Overlay get left behind after the selection is moved.

kép



- [x] Text and overlay of Profile Load appears blurred.

kép



- [x] Control Surface rotation breaks Element Name Overlay in 90° and 270° modes.

kép

- [x] User interaction sometimes makes the Preset Load Overlay disappear. Sometimes user interaction doesn't make it disappear.


## UX considerations

- Design intent when interacting with the module while Preset load is active currently not clear. User interaction while Preset Load Overlay is active after a preset has been selected could activate the loading onto that Element for workflow optimization purposes.

- The way the firmware info overhangs the selection overlay hurts readability. It should either envelop the whole text section or not overlay the text itself to keep this important text always readable.

kép

elsoazemelet:
UX considerations were discussed with @SukuWc. The mentioned bugs were fixed.

Q: "User interaction sometimes makes the Preset Load Overlay disappear. Sometimes user interaction doesn't make it disappear."
A: The any overlays should disappear when for user inputs. User Inputs are selecting a different (!) event or control element.

Q: "Design intent when interacting with the module while Preset load is active currently not clear. User interaction while Preset Load Overlay is active after a preset has been selected could activate the loading onto that Element for workflow optimization purposes."
A: Would be nice, no decision was on this subject. Might need a more fleshed out general solution for handling user input during preset/profile load.

Q: "The way the firmware info overhangs the selection overlay hurts readability. It should either envelop the whole text section or not overlay the text itself to keep this important text always readable."
A: This was rejected, since this would break the displayed design. So between readibility and design, the design was preferred.

PR #503, opened 2023-11-27
Add support for action blocks provided by plugins

danim1130:
The goal of this PR is to provide packages with the ability to add custom action buttons to the action picker component, increasing the functionality of packages.

Test with the following package: https://github.com/intechstudio/plugin-sample-action

- [ ] Package can be loaded into Editor
- [ ] A static plugin action appears when the package is loaded
- [ ] New actions can be added by pressing the add button
- [ ] Pressing any new action on the picker adds a new action to the selected component
- [ ] Any valid LUA code can be written into the text area
- [ ] Storing the changes, then unplugging and plugging in the module still shows the package action
- [ ] Unloading the package removes every action


src/renderer/main/panels/configuration/Configuration.store.js
SukuWc, line 1: Please remove logs if not needed

src/renderer/config-blocks/package/Package.svelte
SukuWc, line 1: Let's review this together! (Maybe today)

SukuWc:
Please also fix formatting!

PR #516, opened 2023-12-12
Fixed local variable suggestions 

### LINKED ISSUE #461: Config manager based local variable suggestions with proper scoping
Provide only valid local suggestions:

Image

elsoazemelet:
Closes #461 

Local suggestions are now reworked.
Previously the anytime a suggestions panel appears for an input field of an ActionBlock, all the defined local variables are displayed for selection. This is done regardless of whether it is defined in that context or not.

Examples when suggestions are showed, but should not (also see the original ticket for image):
midi block (a is suggested)
local.a = 3

if block
   local.b = 3
else block
   midi block (b is suggested)
end block

This behaviour is now fixed, and the suggestions are displayed according to its availability in that context, similarly to if it was written in code.

### DIFF
--- a/src/renderer/main/ErrorConsole.svelte
+++ b/src/renderer/main/ErrorConsole.svelte
@@ -92,6 +92,7 @@
       text = "Ctrl + Shift + R";
     }
 
+    /*
     window.electron
       .fetchUrlJSON(configuration.NOTIFICATION_JSON_URL)
       .then((data) => {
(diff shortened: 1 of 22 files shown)


narayb:
Confirmed works. Should be ok.

src/renderer/main/ErrorConsole.svelte
SukuWc, line 1: Why is this commented out?

SukuWc (commented):
ErrorConsole.svelte? Why

PR #519, opened 2023-12-12
Various UI improvements

elsoazemelet:
Added UI features:

- [x] When selecting the System Element on a module a PCB panel is displayed underneath. The graphic is a crude prototype, review if it is the right direction for further distinguishing the System Element 
- [x] UI Event / System Event selector is removed, since System Events now a clickable element on the module
- [x] When selecting a module, all of its control elements (including System Events) are now selectable from the Configuration Menu via a select list
- [x] Minor UI improvements on the Configuration panels top part (Control Name, Control Name Overlay, etc.)


narayb:
The PCB should use a visual depiction of our own pcb design. This is kind of difficult to evaluate like this. The concept itself fits the 'under the hood' concept of the system event.

Event 255 should not be displayed, as it's not always called as that in Editor syntax. For example timers call the system event the index of the last control element+1 (e.g. on the EN16 the system element timer index is 16).

Visually the Element Selection list/dropdown does not fit the current configuration UI.

elsoazemelet:
- [x] Hide PCB option in developer settings
- [x] Radio buttons should implement melt

PR #523, opened 2023-12-13
Callbacks are reworked to Promises

### LINKED ISSUE #482: Promise based syncronization with grid (instead of callbacks)
Rework functions using callbacks in instructions.js to promised based then/catch.

elsoazemelet:
Closes #482


elsoazemelet:
- [x] intstructions success/fail callback body should be in a then and catch in promise style

Issue #447, opened 2023-10-04
In Editor 1.2.44 autocomplete doesn't react to 'Enter' key in the Locals block textbox


narayb:
Enter should confirm selection of function-autocomplete in all cases.

elsoazemelet:
Investigate if possible with the currently used library

PR #525, opened 2023-12-15
Fixed text clipping for LineEditor and suggestion accepting by pressing Enter

### LINKED ISSUE #444: 1.2.44 Editor long textboxes cuts off text after a while when textbox is stretched
image

As above image shows.

elsoazemelet:
Closes #444 and #447


narayb:
Both are fixed. Hooray!

PR #527, opened 2023-12-18
Small bugfixes for stable

elsoazemelet:
Fixed issues:
- [ ] Inserting IF condition block as first element, then inserting ELSEIF block into it is not possible
- [ ] Inserting ELSE between ELSEIF and ELSE is possible.
- [ ] ELSEIF block is always syntax correct
- [ ] Changing event tracking from event to element loses selected event after user input (no event is selected in config lists event selector)


kkerti:
Checked, looks ok to me.

PR #532, opened 2023-12-20
Added visual feedback for Profile/Preset load

### LINKED ISSUE #392: Load Profile/preset overlay lacks feedback for successful upload
Original idea was that multiple modules can have the same profile and in that case it's good to keep the profile overlay present on the modules.

When using a single module, after "Load Profile" the overlay stays and it feels like the operation did not go through. 

What if:
1. Select profile to upload
2. Show overlay for "Load Profile"
3. After loading to a module - and the load is successful - remove the overlay from that specific module

elsoazemelet:
Closes #392 

Proposed first hand design solution can be observed when loading a preset/profile to a module, and/or changing selected preset/profile.

Check the proposed feedback design in terms of UX.

### DIFF
--- a/src/renderer/main/grid-layout/grid-modules/Device.svelte
+++ b/src/renderer/main/grid-layout/grid-modules/Device.svelte
@@ -90,19 +90,36 @@
   }
 
   function handlePresetLoad(e) {
-    const { elementNumber } = e.detail;
-    configManager.loadPreset({
-      x: device.dx,
-      y: device.dy,
-      element: elementNumber,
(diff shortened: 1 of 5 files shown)


narayb:
Text in front of transparent BG is hard to read. Green transparent box in front of a blue transparent box in front of a black background is also not helping the black text you should be able to read well.

Otherwise works ok, I prefer the icons over the text.

SukuWc:
ok, accepted

SukuWc (approved):

PR #536, opened 2023-12-27
Tooling update

kkerti:
## New update to GitHub Actions
Upload and download artifacts are updated to `v4`. Now the build artifacts are uploaded/downlaoded separately, no need to wait for all jobs to finish to begin testing. [Read more here.](https://github.com/actions/upload-artifact?tab=readme-ov-file#v4---whats-new)

## MacOS notarytool requires teamId
Although uploaded to github secrets and already added to `notarize.js`, the `APPLE_TEAM_ID` was not used in the action scripts. If I remember correctly, I could not use it before as the notarization server was not working when team id was present. Since it's now ultimately required, added back to the scripts.

## Quit app resulting in crash
This branch uses one of the most recent stable branch. Alpha and nightly builds done on this branch don't have this issue anymore. I suspect that the package tab crash or some other background process crashing resulted in this error, which has been fixed in the meantime by @danim1130.

### DIFF
--- a/.github/workflows/alpha-matrix.yml
+++ b/.github/workflows/alpha-matrix.yml
@@ -54,7 +54,7 @@ jobs:
           node-version: "18.16.1"
 
       - name: Download profile cloud artifact
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v4
         with:
           name: profile-cloud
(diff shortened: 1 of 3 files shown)


SukuWc:
Looks good, thank you for your contribution!

PR #537, opened 2023-12-28
 Fixed element overwrite change detection (Part 2)

SukuWc:
**Previous PR merged due to release pressure! One identified issue still needs to be fixed:**

@kkerti  Tested with 1x EF44 + 1x PBF4, side by side.

Scenarios tested:
- [x] change control element init color, copy all, overwrite to other control elements (fader on pbf4)
- [x] change potmeter event (second event for triggering active change), copy all, overwrite to others (fader on pbf4)
- [x] move changed configuration from pbf4 to ef44, see appropiate events showing active changes
- [x] create element preset, no changes triggered
- [x] create module profile, no changes triggered
- [x] load element preset to appropriate element and see the active changes properly trigger
- [ ] after element preset is loaded, **THEN cleared**, new preset loads are not detected correctly

### DIFF
--- a/src/renderer/main/panels/configuration/Configuration.store.js
+++ b/src/renderer/main/panels/configuration/Configuration.store.js
@@ -463,6 +463,7 @@ function create_configuration_manager() {
     });
   });
 
+  // TODO: fails to show overlay after clear
   function loadPreset({ x, y, element, preset }) {
     return new Promise((resolve, reject) => {
       const callback = () => {
--- a/src/renderer/main/panels/configuration/Configuration.store.js
+++ b/src/renderer/main/panels/configuration/Configuration.store.js
@@ -426,7 +426,9 @@ export class ConfigTarget {
 export const configManager = create_configuration_manager();
 
 function create_configuration_manager() {
-  let store = writable(new ConfigList());
+  let store = derived(user_input, (ui) => {
+    return createConfigListFrom(ui);
+  });
 
   function createConfigListFrom(ui) {
     const target = ConfigTarget.createFrom({ userInput: ui });
@@ -456,13 +458,6 @@ function create_configuration_manager() {
     return list;
   }
 
-  user_input.subscribe((ui) => {
-    update((store) => {
-      return createConfigListFrom(ui);
-    });
-  });
-
-  // TODO: fails to show overlay after clear
   function loadPreset({ x, y, element, preset }) {
     return new Promise((resolve, reject) => {
       const callback = () => {
--- a/src/renderer/main/panels/configuration/Configuration.store.js
+++ b/src/renderer/main/panels/configuration/Configuration.store.js
@@ -443,7 +443,9 @@ function create_configuration_manager() {
       const callback = () => {
         runtime.element_preset_load(x, y, element, preset).then(() => {
           const ui = get(user_input);
-          const list = createConfigListFrom(ui);
+          const target = ConfigTarget.createFrom({ userInput: ui });
+          const list = ConfigList.createFromTarget(target);
+
           setOverride(list);
           resolve();
         });
@@ -471,7 +473,8 @@ function create_configuration_manager() {
       const callback = () => {
         runtime.whole_page_overwrite(x, y, profile).then(() => {
           const ui = get(user_input);
-          const list = createConfigListFrom(ui);
+          const target = ConfigTarget.createFrom({ userInput: ui });
+          const list = ConfigList.createFromTarget(target);
           setOverride(list);
           resolve();
         });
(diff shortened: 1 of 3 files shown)


SukuWc:
Peproduction steps provided by @kkerti :

Expanded steps to reproduce:
1. Change a control element's init color to something clearly visible
2. Create an element preset from it
3. Load this element preset to other control elements
4. Up until this point, active changes are working okay
5. **Clear** the module!
6. Reupload the element preset to control elements
7. **Active changes are not triggered anymore!**

Please fix.

https://github.com/intechstudio/grid-editor/assets/47832952/ae5bf9ab-ef76-40aa-ae3e-35f2985f7311

elsoazemelet:
> Peproduction steps provided by @kkerti :
> 
> Expanded steps to reproduce:
> 
> 1. Change a control element's init color to something clearly visible
> 2. Create an element preset from it
> 3. Load this element preset to other control elements
> 4. Up until this point, active changes are working okay
> 5. **Clear** the module!
> 6. Reupload the element preset to control elements
> 7. **Active changes are not triggered anymore!**
> 
> Please fix.
> 
>  prob.mp4

Fixed in the latest commit, it is under testing until tomorrow.

src/renderer/main/panels/configuration/Configuration.store.js
```diff
@@ -458,7 +457,7 @@ function create_configuration_manager() {
   }
 
   user_input.subscribe((ui) => {
-    store.update((store) => {
+    update((store) => {
```
kkerti: This update overwrites the standard store operation as the current code suggests. Such overwrite wrapping is totally doable, but I think it's prone to user errors, specially when revisiting code. The direct `store.update(...)` and `update(...)` is easy to miss.

One possible indicator could be that we say `alteredUpdate(...)` instead of simply `update(...)`. Same improved naming scheme could applied to `set` function as well.

In this case the `setIndentation()` is required after `set` and `update`, so maybe we should make `setAndIndentate` and `updateAndIndendate`.

My point is, looking at this code, I can totally understand why it was easy to miss the specific set/update command required.

src/renderer/main/panels/configuration/Configuration.store.js
```diff
@@ -458,7 +457,7 @@ function create_configuration_manager() {
   }
 
   user_input.subscribe((ui) => {
```
kkerti: Creating a store subscription within a store can result in unexpected store triggers, such subscriptions are considered bad practice in "hot code path" and **doesn't have unsubscription** called anywhere.. Code is hard to understand to me, if the `createConfigListFrom(ui)` is generated in a *reactive way* within a store itself (which should be first called somewhere in a UI component), or it's triggered / called / updated externally.

I think it's worth discovering how the `user_input` subscription could be handled differently here, if we should create a `derived()` svelte store or use some other practice to create "clever svelte objects".

kkerti (commented):
I left some suggestion worth thinking about here.

kkerti:
@elsoazemelet I have checked the latest build, the described bug seems to be gone. Thank you.

src/renderer/main/panels/configuration/Configuration.store.js
elsoazemelet: Fixed. Ambiguous naming and function usage was resolved by using better function naming.

src/renderer/main/panels/configuration/Configuration.store.js
elsoazemelet: Resolved. Subscription is reworked into a derived store.

src/renderer/main/panels/configuration/Configuration.store.js
```diff
@@ -468,7 +468,9 @@ function create_configuration_manager() {
       const callback = () => {
         runtime.element_preset_load(x, y, element, preset).then(() => {
           const ui = get(user_input);
-          set(createConfigListFrom(ui));
+          let list = createConfigListFrom(ui);
+          list = afterUpdate(list);
```
kkerti: Sorry to be a pain in the ass, but `afterUpdate` could be confused with a svelte internal function! Please find a different name.

Otherwise thanks for the refactor, it seems better to me!

https://svelte.dev/docs/svelte#afterupdate

src/renderer/main/panels/configuration/Configuration.store.js
```diff
@@ -426,7 +426,9 @@ export class ConfigTarget {
 export const configManager = create_configuration_manager();
 
 function create_configuration_manager() {
-  let store = writable(new ConfigList());
+  let store = derived(user_input, (ui) => {
```
kkerti: You can't simply slap a `derived()` store in place of a `writable()` store and expect it to work. 

By default, a derived store's value is updated based on the input store's values, first argument of the `derived()` svelte store.

**Please revert this code change.** And return to it's discussion at a future date.

The below update came up the moment I tried to apply a change within Editor with Grid module connected. Did you try your code change before push?

image

PR #540, opened 2023-12-29
Fixed QA feedback of visible active changes during preset load

elsoazemelet:
Fixes QA issues of #392 

Active changes are visible during preset load, similiar to profile load now.

image


kkerti:
### Initial impression
When a preset is loaded the first time on the control element, the *preset loaded!* feedback visually interferes with the active change overlay.
image
### Conclusion ✅
There is room to improve on the above mentioned looks, BUT overall it's much better.

PR #546, opened 2024-01-05
ActionBlock information is cleaned up and is made to be typed

### LINKED ISSUE #460: Integrate information into configManager and make it typed
Create typed interface for information for all action blocks, and integrate it into config manager after making it typed too.
Remove redundant fields

elsoazemelet:
Closes #460 

- ActionBlocks information field is now throwing error when adding not defined properties to it. 
- The property list is defined in typescript in a separate file.
- When a non-optional property is omitted an error is thrown.
- All these thrown errors are only visible during editing the code in VS Code.

### DIFF
--- a/src/renderer/config-blocks/Comment.svelte
+++ b/src/renderer/config-blocks/Comment.svelte
@@ -1,17 +1,16 @@
-<script context="module">
+<script lang="ts" context="module">
+  import type { ActionBlockInformation } from "./ActionBlockInformation.ts";
   // Component for the untoggled "header" of the component
   import RegularActionBlockFace from "./headers/RegularActionBlockFace.svelte";
   export const header = RegularActionBlockFace;
 
   // config descriptor parameters
-  export const information = {
+  export const information: ActionBlockInformation = {
     short: "c",
     name: "Comment",
     rendering: "standard",
-    toggleable: false,
     category: "code",
-    desc: "Comment Block",
-    blockTitle: "Comment Block",
+    displayName: "Comment Block",
     defaultLua: "--[[This Is A Comment]]",
     icon: `
     <span class="block w-full text-black text-center italic font-gt-pressura">--</span>
@@ -33,7 +32,7 @@
   import AtomicInput from "../main/user-interface/AtomicInput.svelte";
   import { Validator } from "./_validators";
 
-  export let config = "";
+  export let config;
   export let index;
 
   const dispatch = createEventDispatcher();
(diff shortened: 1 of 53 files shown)


src/renderer/config-blocks/Comment.svelte
```diff
@@ -1,17 +1,16 @@
...
   import RegularActionBlockFace from "./headers/RegularActionBlockFace.svelte";
   export const header = RegularActionBlockFace;
 
   // config descriptor parameters
-  export const information = {
+  export const information: ActionBlockInformation = {
     short: "c",
     name: "Comment",
     rendering: "standard",
-    toggleable: false,
```
SukuWc, line 11: why is toggleable attribute removed? Comment block should always render opened, ready for editing!

src/renderer/config-blocks/ElementName.svelte
```diff
@@ -1,17 +1,16 @@
...
   import RegularActionBlockFace from "./headers/RegularActionBlockFace.svelte";
   export const header = RegularActionBlockFace;
 
   // config descriptor parameters
-  export const information = {
+  export const information: ActionBlockInformation = {
     short: "sn",
     name: "ElementName",
-    rendering: "standard", //'hidden'
-    toggleable: false,
```
SukuWc, line 11: why is toggleable attribute removed? ElementName block should always render opened, ready for editing!

SukuWc (changes requested):
eventtype and toggleable parameters must not be removed! Toggleable is used for comment and elementname actionblocks to make sure they are always rendered open. Eventtype is used by the actionpicker to only provide compatible special actionblocks. (When they are being added to one of the events specified in the information.eventtype array!)

src/renderer/config-blocks/ElementName.svelte
elsoazemelet: The property toggleable was present in ElementNames.svelte twice with different values. The unnecessary and incorrect one (above) was removed.

The toggleable property is currently used for defining that whether an action block has an openable body or not. Setting toggleable to false causes the action block to always "show it's face", e.g.: If block, For loop block.

An always open ElementName and CodeBlock is doable, similiarly to the If block and For loop.

src/renderer/config-blocks/Comment.svelte
elsoazemelet: The property toggleable was present in ElementNames.svelte twice with different values. The unnecessary and incorrect one (above) was removed.

The toggleable property is currently used for defining that whether an action block has an openable body or not. Setting toggleable to false causes the action block to always "show it's face", e.g.: If block, For loop block.

An always open ElementName and CodeBlock is doable, similiarly to the If block and For loop.

elsoazemelet: Eventtype is removed since the behaviour of the filtering of unaddable action blocks is defined in ActionPicker.svelte.
The filtering is using a variable (eventString) that defines the type of the event, and the config blocks that are unaddable on that event are manually filtered out.

SukuWc (approved):

PR #547, opened 2024-01-07
Increased timeout when writeBuffer is busy to avoid module drops

elsoazemelet:
Increased Timeout for modules. 

**Until first review:**

- Comments are left in for checking the intervals between heartbeats (displayed in ms) to more easily settle on a different heartbeat interval.
- The Timeout was 300ms
- The usual interval between two heartbeats are around ~250ms
- In developement enviroment, the modules are dropping during clear with 1600ms elapsing between heartbeats. 2700ms were present for clearing and storing heavier profile cloud profiles.
- Increased Tiemout to 5000ms to ensure that no drops should occur, eliminating other possibilities for module drop.

**TODO:**

- [ ] After review remove console.logs
- [ ] Set a proper timeout


SukuWc:
Test with nightly firmware

PR #556, opened 2024-01-19
Handle user errors in login panel

### LINKED ISSUE #557: User not found causes runtime error
1. User not found causes runtime error
2. Sign in using google opens dev profile cloud /autorize root in browser even if production cloud is selected in preferences

Image

### LINKED ISSUE #565: Login panel causes runtime error
Image

danim1130:
Handle login errors if user inputs incorrect email or password data.

- [ ] UserLogin modal shows error if email or password is incorrect
- [ ] The error disappears if either field changes

### DIFF
--- a/src/renderer/lib/auth.store.ts
+++ b/src/renderer/lib/auth.store.ts
@@ -10,9 +10,7 @@ import {
   signInWithCredential,
   signOut,
 } from "firebase/auth";
-import {
-  FirebaseError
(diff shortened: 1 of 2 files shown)


SukuWc (changes requested):
Please update auth.store.ts and UserLogin.svelte to use .catch syntax instead of await + try catch block!

Issue #566, opened 2024-01-24
Stable fixes: Virual module category


elsoazemelet:
- [x] Select dropdown mouse event must be fixed 
- [x] Object.keys(CEEAT).find((key) => CEEAT[key].value == e.type)
- [x] HWCFG for virual modules from grid-protocol.js
- [x] actionstring for virual modules from grid-protocol.js

PR #569, opened 2024-01-24
🚩PR: Various QA fixes of stable

### LINKED ISSUE #567: Stable fixes: Promise category
- [x] Promise based pageStore top level interface
- [x] Module disconnect removes items from writebuffer.

elsoazemelet:
Closes #567 and #566 and #568

The issues fixes and features implemented:

- [x] In virtual module mode the clear/discard/store functionalities should work as with normal modules. Discard resets the module configuration to the last stored state, store stores the current state, and clear resets to the default configuration. 
- [x] Virtual Modules can be added properly on its modal dialog, without the dialog closing when selecting module type from the dropdown list.
- [x] Virtual modules should have their proper default configuration
- [x] In normal module mode, the discard, store, and clear should feedback and work as intended. (check logger messages)
- [x] Disconnecting modules during running processes (e.g.: during page store, profile load, etc.) should not break editor.

### DIFF
--- a/src/renderer/runtime/engine.store.ts
+++ b/src/renderer/runtime/engine.store.ts
@@ -128,7 +128,9 @@ function createWriteBuffer() {
     timeout: number
   ): Promise<GridResponse> {
     waiter = new ResponseWaiter(bufferElement, timeout);
-    return await waiter.waitResponse();
+    const response = await waiter.waitResponse();
+    waiter = undefined;
+    return response;
   }
 
   async function sleep(time: number) {
@@ -157,17 +159,20 @@ function createWriteBuffer() {
         });
 
         do {
-          const { id } = await sendDataToGrid(incoming.descr);
-          if (
-            incoming.responseRequired &&
-            incoming.filter !== undefined &&
-            incoming.filter.class_parameters !== undefined &&
-            incoming.filter.class_parameters["LASTHEADER"] !== undefined
-          ) {
-            incoming.filter.class_parameters["LASTHEADER"] = id;
+          let id = undefined;
+          try {
+            const result = await sendDataToGrid(incoming.descr);
+            id = result.id;
+          } catch (e) {
+            reject(e);
+            return;
           }
 
           if (incoming.responseRequired === true) {
+            const { class_parameters } = incoming.filter || {};
+            if (class_parameters?.LASTHEADER !== undefined) {
+              class_parameters.LASTHEADER = id;
+            }
             const timeout = incoming.responseTimeout ?? 1000;
             const result = await waitResponseFromGrid(incoming, timeout);
             switch (result.status) {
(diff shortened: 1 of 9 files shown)


narayb:
Loading a (possibly faulty) button preset on a virtual TEK2 produces error:
kép

src/renderer/runtime/engine.store.ts
```diff
@@ -157,17 +159,20 @@ function createWriteBuffer() {
...
-          const { id } = await sendDataToGrid(incoming.descr);
-          if (
-            incoming.responseRequired &&
-            incoming.filter !== undefined &&
-            incoming.filter.class_parameters !== undefined &&
-            incoming.filter.class_parameters["LASTHEADER"] !== undefined
-          ) {
-            incoming.filter.class_parameters["LASTHEADER"] = id;
+          let id = undefined;
+          try {
```
SukuWc: Set id in .then section and  reject in .catch section. TryCatch block is not needed, just await sendDataTo.Grid(incoming.descr).then(e){...}.catch(e){reject(e)}

src/renderer/serialport/serialport.js
SukuWc, line 1: These changes are welcome!

SukuWc (changes requested):

src/renderer/runtime/engine.store.ts
elsoazemelet: Fixed. I reorganized this part of the code for the better understanding of the process, including that this try/catch mechanism was promisified. Please validate if it is the correct way! Thank you for the feedback!

elsoazemelet:
> Loading a (possibly faulty) button preset on a virtual TEK2 produces error: kép

@narayb the reason of the error is that this button preset has 4 event configurations on it. Possibly the reason is because it is actually an encoder preset? Not sure, but fixed so the unknown event is not transferred to grid. @SukuWc  is this valid? If I remember correctly, this feature was implemented before, but I accidentally cut it out during refactor?

narayb:
`Disconnecting modules during running processes (e.g.: during page store, profile load, etc.) should not break editor.`

The above works well, but also allows for the loading of half-profiles which result in broken functionality on the Grid side, without much indication of this happening on the Editor side.

Logger should show that an unexpected disconnect occurred during store and for a successful store, the user should redo the process.

This way we warn the user on the software side as well.

elsoazemelet:
> `Disconnecting modules during running processes (e.g.: during page store, profile load, etc.) should not break editor.`
> 
> The above works well, but also allows for the loading of half-profiles which result in broken functionality on the Grid side, without much indication of this happening on the Editor side.
> 
> Logger should show that an unexpected disconnect occurred during store and for a successful store, the user should redo the process.
> 
> This way we warn the user on the software side as well.

Fixed the feedback message. I do believe that there was already a logging of the error of store/discard/clear when disconnecting, but the now changed feedback message is more prominent.

narayb:
Feedback message tested and working!

PR #571, opened 2024-01-24
Mock Preload for Browser Mode

kkerti:
### What is this?
Proof of concept for mocking electron functions exposed from `src/electron/preload.ts` with a single `window` object overwrite.

### To do...
Test everything which can be clicked with a mouse and continue adding mock values under `src/renderer/preload-window-config.ts`. The preferences panel's developer settings is broken for example.


SukuWc:
Works well in browsers that support serial, but navigator.serial is undefined in Firefox. Please handle this gracefully.

Uncaught TypeError: can't access property "addEventListener", navigator.serial is undefined

kkerti:
@SukuWc analytics with mixpanel is using now the regular editor configuration, it should be revisited before release!

SukuWc:
TODO comments defined in preload-window-config.ts! (configuration.json and buildVariables.json must be properly loaded!)

kkerti:
@SukuWc updated.

PR #573, opened 2024-01-26
🚩PR: Fixed runtime error for dropdown select list

### LINKED ISSUE #560: Improper use of MeltSelect can cause runtime error
## MeltSelect
- make export let options; fully typed
## Fix Element Select panel

Image

### LINKED ISSUE #582: Virutal module bug: Element selection dropdown in configuration panel causes runtime error
Tested on current stable branch!

Image

elsoazemelet:
Closes #560 

The fixed issue was the instable state management of MeltSelect.svelte. Fixed the selection state management logic.

## Testing

The stresstesting of the select list is probably the way to go.

This component is used as a dropdown list in the following parts of editor:

- [x] In the tracker options (track events, elements, or off)
- [x] In the preferences menu
- [x] In the configuration menu, with the control element selection
- [x] In the virtual module adder modal, with the type selection

Closes #582 

Test dropdown for control element selection on Configuration panel in virtual mode too. See original issue of #582

### DIFF
--- a/src/renderer/main/panels/preferences/MeltSelect.svelte
+++ b/src/renderer/main/panels/preferences/MeltSelect.svelte
@@ -33,6 +33,8 @@
     handleTargetChange();
   }
 
+  $: console.log(target);
+
   function handleTargetChange() {
     if ($selected.value === target) {
       return;
--- a/src/renderer/main/panels/preferences/MeltSelect.svelte
+++ b/src/renderer/main/panels/preferences/MeltSelect.svelte
@@ -8,7 +8,7 @@
 
   function getDefaultSelected() {
     const obj = options.find((e: SelectOption) => e.value === target);
-    return { label: obj.title, value: obj.value };
+    return { label: obj?.title, value: obj?.value };
   }
 
   const {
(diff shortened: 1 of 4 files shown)


narayb:
Seemingly works as intended.

src/renderer/main/panels/preferences/MeltSelect.svelte
```diff
@@ -1,9 +1,15 @@
...
-  export let target;
+  export let options: SelectOption[];
+  export let target: any;
   export let size: "auto" | "full" = "auto";
-  let oldTarget;
+
+  type SelectOption = { title: string; value: any };
+
+  function getDefaultSelected() {
+    const obj = options.find((e: SelectOption) => e.value === target);
```
elsoazemelet, line 10: Make it possible to accept invalid or undefined default target => Empty selectbox

src/renderer/main/panels/preferences/MeltSelect.svelte
```diff
@@ -16,51 +22,16 @@
...
-        });
-      }
+  $: handleSelectionChange($selected);
 
-      if (target !== $selected.value) {
-        oldTarget = target = $selected.value;
-      }
+  function handleSelectionChange(selected: { label: string; value: any }) {
+    if (selected === target) {
+      return;
```
elsoazemelet, line 47: Use case:
NEVER update target if no actual selection change happened
If target is updated from the outside (correctly) change selection, (target should not change twice, in this case again from the inside)
If target is changed from the inside, propagate change to the outside.

PR #576, opened 2024-01-29
🚩PR: Action picker resizes to window height

### LINKED ISSUE #559: Add action panel clips instead of showing scrollbar when editor window is not tall enough
Image

elsoazemelet:
Closes #559 

Details are in the original ticket.


narayb:
Neat!

PR #585, opened 2024-02-01
🚩PR: Refactored GridProtocol

### LINKED ISSUE #558: grid-protocol.js must expose HWCFGs, default configs, event type enums, element type enums

elsoazemelet:
Closes #558 

- grid-protocol.js was changed to grid-protocol.ts
- Necessary types were created, and the enums were moved from virtual-engine.ts to grid-protocol.ts
- These enums and types are used through grid-protocol.ts now instead of implicit types and enum values are used instead of string values
- The exposed lookup-table style properties (elementEvents, moduleElements) are removed, and functions are used instead. 
- The usage of these lookup-tables are refactored to use the new functions
- grid is now an exported instance of a GridProtocol class, instead of the default export of an inline object
- The usage of grid-protocol.js, and the direct use of grid-protocol JSON is refactored out, now everything uses the new grid-protocol.ts


SukuWc:
please create interface for getting values from the grid.properties. Properties should be private! No direct access to member variables!

PR #598, opened 2024-02-11
🚩PR: Heartbeat is now sent out whenever is possible to grid

### LINKED ISSUE #597: Heartbeat message must preempt all messages in WriteBuffer (even pending ResponseRequired ones)

elsoazemelet:
Closes #597 

Easy fix of serial port communication (engine) logic, to accomodate the handling of sending out immediate heartbeate messages to grid.

Now heartbeat messages are sent out whenever the serial port is available.

This should result in combination with the newsest nightly firmware so that the grid modules should not drop during any operations.

### DIFF
--- a/src/renderer/config-blocks/ActionBlockInformation.ts
+++ b/src/renderer/config-blocks/ActionBlockInformation.ts
@@ -5,11 +5,7 @@ export interface LuaScript {
   script: string;
 }
 
-type EnsureNonOptional<T> = {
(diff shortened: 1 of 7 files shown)


elsoazemelet:
TODO:

- [x] Add sendImmediate property to BufferElement with documentation
- [x] Move buffer element into engine
- [x] Rethink sendImeddiate condition, if possible, no parenthesises
- [x] Rename executeFirst/Last back to addFirst/Last

PR #599, opened 2024-02-11
🚩PR: Fixed error of unhandled exception when scheduler breaks

### LINKED ISSUE #580: Stable Nightly Editor freezes and module disconnects when sending too many messages with a certain profile
## Steps to reproduce
- load profile Seque 1.3 on an EN16 (tested on D51, FW v1.2.29)
- turn the speed of the sequencer up by rotating the bottom right encoder clockwise
- module will disconnect from Editor and throw that error

Pasted image 20240130151028

(on release version Editor 1.2.48 the same steps only produce disconnects with instant reconnects)

elsoazemelet:
Closes #580 

The detailed description of the fixed bug can be found in the linked ticket.

**IMPORTANT:** The module during fast cycling repeats actually disconnects and reconnects, since the rate of received heartbeats slow down. This is also an issue.

The fix contains the elimination of the unhandled exception during the repeated and immediate drop and re-connection of a module. The unhandled exception should not appear now when reproducing the bug, only the (normal) error message in the console. 

The editor now should not freeze when this occure, and should be operational when the sequencer is stopped, the cycle rate is decreased or when the module is re-connected physically.

### DIFF
--- a/src/renderer/main/grid-layout/grid-modules/overlays/PresetLoadOverlay.svelte
+++ b/src/renderer/main/grid-layout/grid-modules/overlays/PresetLoadOverlay.svelte
@@ -60,7 +60,7 @@
     const { dx, dy } = obj;
     const ui = get(user_input);
 
-    const target = new ConfigTarget({
+    const target = ConfigTarget.create({
       device: { dx: dx, dy: dy },
       page: ui.pagenumber,
       element: elementNumber,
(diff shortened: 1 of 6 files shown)


narayb:
Tested, error message is gone. Module stays connected. Profile is still faulty, but that's not our problem.

PR #605, opened 2024-02-16
Every button is now MoltenPushButton

elsoazemelet:
All buttons are now using the component MoltenPushButton.
The design of all buttons are now unified with three introduced styles:

- Normal: A black one, using the design language of Preferences
- Accpet: A green one, that is useing the design of the store button
- Outlined: A transparent one with green border, using the design of the troubleshoot button on the module connect dialog

Interesting features:

- The action block operations (copy, paste, cut, etc.) buttons are now using this new MoltenPushButton as a base.
- The confirm style tooltip used at the clear button to confirm the clear is using now this new MoltenPushButton.

**Actions to do:**

Review editor, and determine further advancements in design unification


narayb:
Pasted image 20240220155358
Visibility is still horrible in front of dark backgrounds. 

Pasted image 20240220155839
They look okay when the background is lighter than the button background...


Pasted image 20240220155914
...but all that goes out the window when there is no light background.



All in all, the **buttons look nice**, but **usability suffers with every change we make to these buttons** without changing the contrast of either the backgrounds or the buttons or their borders.

Please use this  now complete technology change to update the design and make the buttons pop from their backgrounds again.

PR #620, opened 2024-02-26
SUKU graph based debug option added to developer options

SukuWc:
This feature is used to see if messages were dropped during transfer. Each module stores the last message id that it has sent out. When creating a new packet, the module increments this variable. This is a uint8_t, values are between 0...255. Ofter 255 the next message ID will be 0. The first graph displays the last message ID of the given module, the second graph displays the difference between the last message ID and the current one. This difference should always be 1 indicating that every single message arrived and they arrived in the order they were created. In 4+ module setups this is non-trivial. Currently occasional packet drops do happen in normal use, and packet drop always happens during page change.  This editor feature is meant to aid the development and verification of the grid firmware!

- Enable this feature in developer options
- See result in Debug side-panel

image


SukuWc:
During page change there are packaets droppend indicated by the graphs and the Invalid Checksum debugtext messages 
image

PR #624, opened 2024-02-29
🚩PR: Added feature of GridLayout repositioning

### LINKED ISSUE #445: Allow moving the control surface in the layout to different positions
Previously with the `panzoom` library, moving the control surface in the center layout area was possible. This feature is cool to create blog - marketing - other content, where the framing of the configuration can be changed.

elsoazemelet:
Closes #445 

- Now when holding down the control key and click + dragging with the mouse, the displayed modules can be repositioned
- This repositioning is resets when clicking the Reset Grid Layout button on the Tracker
- Check the tooltip for this feature next to the reset button
- When the sticky container (that contains the page changer buttons) is would be outside of the middle panel, then it automatically sticks to the bottom of the screen
- When no module is connected the positioning is reseted to the center
- This feature is not persistent, the restart of the editor should reset the positioning.

### DIFF
--- a/src/renderer/main/MiddlePanelContainer.svelte
+++ b/src/renderer/main/MiddlePanelContainer.svelte
@@ -12,6 +12,7 @@
   import ModuleHangingDialog from "./user-interface/ModuleHangingDialog.svelte";
   import StickyContainer from "./user-interface/StickyContainer.svelte";
   import { onMount } from "svelte";
+  import ControlSurface from "./panels/configuration/components/ControlSurface.svelte";
 
   let logLength = 0;
   let trackerVisible = true;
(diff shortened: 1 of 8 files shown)


narayb:
Reset Grid Layout is not a very descriptive name for the button that resets the _position_ of the whole layout. Preferably is should somehow tell the user about that.

This button received the same highlight colors that all other buttons use now, like the Discard button. I still feel like that their highlighting is way too unnoticeable. Dark button in front of a dark background becoming slightly less dark is not a great way to highlight.

For some reason the tooltip disappears after clicking on the tooltip questionmark. This issue affects all tooltips, made a separate issue about this here: https://github.com/intechstudio/grid-editor/issues/629.

All other features work as described.

PR #625, opened 2024-02-29
🚩PR: Added setting of Disable Animations

### LINKED ISSUE #375: User request: Disable all animations from perferences
Add option to disable all animations in preferences.
Connect this option with the svelte transitions and css transitions.
These are:

- transition-<something> in tailwind
- transition:<something>={...} in svelte
- in/out:<something>={...} in svelte
- anything that uses something from "svelte/transition"
- etc. be creative

elsoazemelet:
Closes #375 

All EDITOR transitions and animations can be disabled from Preferences. This also automatically work without explicitly setting the animations to disabled, if in the OSs settings has the option for preferred reduced motion.

This should automatically work for ALL animations in EDITOR

Examples for the disabled animations:

- Switching up ActionBlocks
- Connecting Modules
- Logger message display (no fade in)


narayb:
Firmware update animation is also enabled, makes it so that no red border appears at all.

Please make it so that a fallback static red border appears when animations are disabled in this case.

elsoazemelet:
- [x] Add 3-way radio button for changing animation (on, off, auto). Display prefer reduced motion value.
- [x] Add red border to FW update modules when animations are disabled

PR #626, opened 2024-03-01
🚩PR: Improved module selection for virtual mode and added instant virtual mode for web version

### LINKED ISSUE #613: Virtual module UX updates
- Create dedicated floating panel for visually selecting from available module types
- npm run web-dev cannot open serial port, so show the newly created floating panel for virtual modules instead of connection panel
- Create option to delete virtual module

elsoazemelet:
Closes #613 

- Improved Virtual Module creation experience. When opening the Virtual Module picker dialog, instead of the previous select list, the desired module can be selected from a grid that contains the visual representations of the available modules.
- Selected Virtual Module now can be changed with a dedicated button. This button is present under the page changer buttons. If the profile/preset load overlay is displayed, then the Close Overlay and Module Selector button is positioned next to each other, and should not interfere with each others function.
- If the editor is opened in the browser (special or developer build may be needed to test this feature) the Titlebar and Welcome dialog is not rendered, and the Virtual Mode is instantly started with the opening of the Virtual Module selector.


narayb:
Features work as described.

## Curious interaction
When you plug in a module with firmware update notification, unplugging the module permanently puts Editor in a state where the "waiting for bootloader to enumerate" banner is displayed.

kép

From this state the only way to access virtual modules again is to reload the application or to FW update the module.

PR #639, opened 2024-03-07
🚩PR: Fixed bug of clipped tooltips on middle panel container 

### LINKED ISSUE #629: Tooltips disappear after clicking, active bug on both release and nightly stable
For some reason each Editor tooltip disappears after clicking on the tooltip questionmark after hovering it.

tooltip

Expected behavior should be that the tooltip stays visible, regardless of clicking until the hover lasts.

elsoazemelet:
Closes #629

- Fixed tooltip box clipping on middle panel container (resize the window, so the tooltip of the Clear button would overlap the right side panel. Now it is not clipped, but displayed properly)
- Fixed original issue (see ticket) of disappearing tooltip when clicking on the Questionmark based tooltip (e.g.: next to Element Name)


elsoazemelet:
- [x] Click type tooltip should remain opened while hovering the tooltip

PR #676, opened 2024-04-10
🚩PR: Design proposal for action button toolbar

### LINKED ISSUE #654: Action toolbar names should all be unique
kép

Change the Copy of Elements to 'Element Copy' etc. so that they do not match.

Remove category names.

kép

Same textname should appear in context menu.

elsoazemelet:
Closes #654 

The original ticket is to give unique names to the action buttons: 

image

This PR however contains a design proposal for these buttons that aims for the following goals:

- The element and action block actions are separated (into two rows)
- All action have a designated color code, that helps to reassure users whether the correct button is hovered. This helps for users to validate that the correct button is selected, that helps to recognize correct patterns quickly. This should cause a quick, seamless and reliable editing experience.
- The hovered buttons function is displayed next to the buttons. This helps for first time users to get to know the function of each actions. Without displaying constantly the name of the actions, it is easier to recognise the position of a button, that further helps pattern recognition. This is especially true, since the compact design rarely if ever should change reactively during window resize
- By not needing to display short action names next to icons, the function of an action button cna be displayed instead of its "name".


narayb:
Compared to the current solution, the new solution doesn't solve the following problems:
- icons are still not recognizable at a glance (there was no change to the icons)
- coloring is not visible when interaction happens, only on hover - this doesn't improve the recognition of operation at a glance
- the removal of operation names worsens UX, since the buttons are still unrecognizable at a glance
- only showing the hovered action names, still makes the user read, but now they have to hover everything before reading them - the removal of names is only possible if buttons are recognizable at a glance to the user without names, not the case right now


Suggestions:
Revert to release version, but add the coloring to the icons when the operation should be available. For example:
- Copy Element and Overwrite Element (when clipboard has an element) should always have color, since one Element is always selected.
- Discard Changes from Element should be colored when there are active changes on that Element.
- When an Action is selected, all Element Operations should lose their color.
- Copy, Cut, Merge, Delete Action operations should get colored when there is at least one Action block selected.
- Paste operation should gain color when there is an Action or Actions on the clipboard.

This would help increase readability of the buttons, but wouldn't remove the names and undo some of the bandaid UX it provides.
Also the Element operations should gain the 'Element' prefix still where applicable when they gain their names back.

PR #687, opened 2024-04-17
🚩PR: Added syntax check preprocessor

### LINKED ISSUE #680: Action block provided syntax check preprocessor function.
Solution A:
- Actionblock provides optional syntax checp preprocessor (string->string) function
Solution B:
- When dispatching script, actionblock also dispatches scriptForSyntaxCheck variant

elsoazemelet:
Closes #680 

Reworked syntax checking process to utilize a syntax preprocessor that is defined by the individual action blocks.
If the preprocessor is not defined, the syntax check is done based on the action blocks script.

This change affects all action blocks, but the significantly impacted action blocks are the following:

- Button Press/Release
- All IF blocks
- All Encoder blocks
- For Loop
- Raw Code

### DIFF
--- a/src/renderer/config-blocks/ActionBlockInformation.ts
+++ b/src/renderer/config-blocks/ActionBlockInformation.ts
@@ -9,9 +9,10 @@ export type ActionBlockInformation = Information;
 
 export class SyntaxPreprocessor {
   constructor(public generatorString: string) {}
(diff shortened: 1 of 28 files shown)


narayb:
Looks ok in tested scenarios, didn't encounter issues.

PR #688, opened 2024-04-17
SUKU endless functions implemented, sriptToSegment now resprects commas

SukuWc:
Make sure to use latest nightly firmware https://github.com/intechstudio/grid-fw/raw/preview/Preview/Firmware/grid_esp32_nightly.uf2


narayb:
Upon Page cycle LEDs don't refresh to the stored value on init event, but stay at 0.

Issue #694, opened 2024-04-24
Keyboard shortcuts for config actions


elsoazemelet:
- [x] Disable element copy/overwrite when one or more action-blocks are selected
- [x] CTRL+C should copy actionblocks, if no actionblocks are selected then it should copy whole element
- [x] CTRL+V should paste actionblocks if actionblocks were the last on cllipboard, otherwise trigger whole element overwrite
- [x] Visually show shortcut triggering of these actions on the buttons!
- [x] Implement reset to default config context menu item for elements
- [x] Multiselect toolbar text should display shortcut
- [ ] Ctrl -> Command on Mac, also test it

PR #705, opened 2024-04-28
🚩PR: Shortcuts are added for action buttons

### LINKED ISSUE #697: Cut doesn't actually Cut, tested on 1.2.52, active on 53 nightly stable as well
cut_bug52

elsoazemelet:
Closes #697  and #694

- Added shortcuts for every toolbar action
- Shortcuts are displayed next to the actions name, left to the toolbar
- Shortcuts take effect when the application is in focus
- Fixed Cut action
- Unified clipboard for Element/ActionBlock copy, now if copying the whole element overwrites the clipboard data of ActionBlocks and vice versa
- If ActionBlocks are selected, the Element Copy is unavailable. This causes that the Ctrl+C shortcut copies whole element if no ActionBlocks are selected, otherwise the selected ActionBlocks
- Added feature of Element Clear, that works similiar to the regular Clear. The difference is that it only resets to the default settings the selected Element.
- Added feature of Element Clear to the Context Menu
- Element Clear is always available, and is not automatically stored so it can be discarded.


kkerti:
On web view, it does not work. I'm aware it's not a primary focus, but it would be a good reassurance for mac os users. Please look up the device on the navigator.

https://www.geeksforgeeks.org/how-to-detect-operating-system-on-the-client-machine-using-javascript/

elsoazemelet:
- Paste seems broken (length check)
- Hotkey in new line + truncate

PR #711, opened 2024-05-05
🚩PR: Added feature of search in ActionBlock picker

### LINKED ISSUE #692: Action pick search bar
Search matches if category or action name substring matches, similar to profile cloud search!

elsoazemelet:
Closes #692 

This PR contains the following features:

- A search bar is added to the action picker
- All partial matches should be displayed (using the ActionBlocks displayed name in the ActionPicker)
- If no matches found, it is displayed accordingly
- The height of the ActionPicker depends on the available ActionBlocks
- The width of the ActionPicker now fixed, and adapts to the screensize


elsoazemelet:
- Auto focus search bar
- Auto select first result
- Enter should paste
- smart search -> PC

kkerti:
Cool! I think it would be better to have a min-height on the action panel even though search is active. Keep the same height upon searching like you open the panel.

PR #712, opened 2024-05-05
🚩PR: Replaced LUA format

### LINKED ISSUE #691: Patch Lua minifier to ignore line comments and block comments

elsoazemelet:
Closes #691 

Editor is reworked to utilize the lua-fmt package instead of the previously used lua-format to fix the following PREVIOUSLY existing (but patched) issues of lua-format:

- Inifinity loop when entering [[ ► no freeze, no infinity loop
- valid syntax of func(a,a,) ► this should be invalid
- everything is valid after a single end; keyword ► syntax check should be working as intended
- watermark is printed ► no watermark

**ALSO this should fix the persisting issue (which was the task) of unintended removal of line and block comments**

All in all, every LUA syntax valid code should pass the syntax check, and all invalid syntax should be detected without issues, and the errors should be displayed correctly.


elsoazemelet:
- Rework minifying
- Make unit tests to test for minifying

elsoazemelet:
- check build pipeline for jest output

PR #723, opened 2024-05-14
Fixed quote replacing

elsoazemelet:
Fixed quote replacing in lua-fmt during the beutify process (formatText).
Checked test case: 
local str = "hello('(d'"


SukuWc:
please add this example to the unit tests to make sure that this never breaks again!

PR #726, opened 2024-05-17
🚩PR: Web build now can use WebSerial API to use physical grid modules

### LINKED ISSUE #719: Implement grid connection in browser build using the WebSerial API

elsoazemelet:
Closes #719 

- Everything should work as previoulsy in the desktop build of Grid Editor
- Added feature of requesting USB usage in web build of Grid Editor
- When USB port access is enabled, physical grid modules can be used as in desktop builds

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "firebase": "^9.18.0",
     "fs-extra": "^10.0.0",
     "googleapis": "^100.0.0",
-    "grid-protocol": "^1.2.36",
+    "grid-protocol": "1.2.36",
     "is-online": "^10.0.0",
     "lua-fmt": "^2.6.0",
     "luamin": "^1.0.4",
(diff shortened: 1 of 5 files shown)


elsoazemelet:
image

kkerti:
Seems to work under chrome, but further testing required. I was able to quickly store a config change on my single connected module. After disconnecting the module and reconnecting, the "module connect" has to be re-enabled. Not sure if feature or bug.

**Opinion:**
The "Connect your modules!" should be opening up only on Chrome only. I think supporting other browsers is a fight not worth to pick. I think the writing on the modal should be something like:
`You can use the browser Editor to configure your modules under Chrome only. Download Chrome *here*. If you are looking forward to explore Editor in virtual mode, start by adding virtual modules.`
The modal should check the running browser type.

Better ux writing should be checked with @narayb.

It is a feature. Currently there is no way to solve the issue of reconnecting after the port was lost automatically.
The fix of runtime error during physical module connection enabling under browser differing from Chromre was implemented. The proper wording is to be selected, I will contact QA!

PR #756, opened 2024-06-20
🚩 PR: Unified action state management

### LINKED ISSUE #748: Copy - Overwrite element does not work
Copy successfully copies events, but overwrite remains disabled.

Suku comment:
- Make sure that context menu enabled state and action icon state is consistent

elsoazemelet:
Closes #748

Unified internal state management for config actions (Copy, Paste, etc.).
After this PR mismatched states between the Toolbar and Context Menu should not happen.

Example for mismatched states:
Overwrite element is enabled in the Context Menu, but not in the Toolbar.


narayb:
In testing, the Discard Element sometimes failed to discard changes on Virtual modules.

Issue could be reproduced by loading the 'Loopy Pro Controller v1.3' Profile on a virtual EN16 and then trying to copy and overwrite encoders. When trying to discard the overwrites, they would remain in all cases active. Only discard all would get rid of these.

PR #758, opened 2024-06-24
🚩 PR: Added action helpers

### LINKED ISSUE #757: Add action helper when special action block is empty
- Display face actionblock that propts the user to add actionblocks, similar to what is displayed when there are no actionblocks on an event
- All special actionblocks should provide this feature
- Text should be custobized based on the special actionblock

Example text: 
- Encoder Left/Right: Add actions that will be triggered when you turn the encoder left
- IF: Add actions that will be triggered "if value>10" condition is met
- ELSE: Add actions that will be triggered "if value>10" condition is NOT met

Image

elsoazemelet:
Closes #757 

- Added feature of helper text for special action blocks (affected action blocks: encoder actions, button action, if conditions, and for loop)
- Reworked Add Action button
- Fixed syntax error border display when the action block that has error is nested
- Added function to turn off this feature in preferences (default value is enabled)

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,5 +18,3 @@ npm run format
 npm i
 npm run web-dev
 ```
-
-© Intech Studio Ltd.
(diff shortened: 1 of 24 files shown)


SukuWc:
Looks good, few improvements before we can merege
image
- Use blue color for adding actions to stay consistent
- Button on the far right of the box makes it awkward to navigate, please mow it to the left or make entire area clickable

narayb:
Functionality is good!

PR #761, opened 2024-06-25
🚩 PR: Added Code Path display for Monaco editor

### LINKED ISSUE #760: Monaco modal should display current edited codeblock path
Display the following path in the top of the modal dialog
PBF4 (0,0) / Element 4 / Init / Code Block #2

- [x] Display module name and X,Y coordinates
- [x] Display element number or element name if available
- [x] Display the event name
- [x] Display the Index of the currently edited codeblock in the confglist

elsoazemelet:
Closes #760 

~Feature is still work in progress...~

Added code block path in Monaco editor. The path changes when the Code Block is moved, or deleted. Code is no longer committable after CodeBlock deletion.


SukuWc:
Design looks good but there are some unhanded cases
- Add two code-block to an event to position 3 and 4
- Open and commit some change to actionblock at position 4 (keep monaco modal open)
- Use drag and drop to change the order of the two actionblocks (keep monaco modal open)
- Commit some other change. The change will be committed to position 4 (expected would be position 3)

Visual:
- maybe add highlighting to show the user visually which block is being edited!

narayb:
Same issues as above. Make it more consistent. When changing edited blocks, the contents of the monaco editor should change, not retain the contents from the previous Code Block.

elsoazemelet:
Fix code commit bug when changing target

PR #765, opened 2024-07-03
🚩 PR: Added partial preset load and element override

### LINKED ISSUE #462: Load button preset to encoder, overlay shows encoders and buttons in different color
Warn the users that it might cause invalid state. Example, encoder color is lost when overriding init.

elsoazemelet:
Closes #462 and https://github.com/intechstudio/grid-protocol/issues/9


elsoazemelet:
- [x] No blinking of Configmanager when loading preset
- [x] No compatibility when preset is broader than the target, only when it is narrower 
- [x] Move element compatibility functions to grid-protocol

PR #768, opened 2024-07-05
🚩 PR: Added remote firmware check

### LINKED ISSUE #766: Firmware update should not update to latest, but to exact version that comes from CMS.
- FW required version is specified in editor
- FW required version can be overwritten from CMS (github static file) similar to notification.json
- File name in repository should be firmware.json

elsoazemelet:
Closes #766 

Now required firmwares can be binded to editor versions in firmware.json. 
If internet connection is not available, or the there is no matching entry in firmware.json, the default values are used in configuration.json


elsoazemelet:
Reset hardcoded editor requirements
CMS > editor requirements

SukuWc:
Firmware download method still only downloads latest firmware. Please refactor firmware.ts:
- [x] firmwareDownload and firmwareNightlyDownload should not exist separately
- [x] firmwareDownload should have following args: targetFolder, product, arch, url
  - targetFolder is fetched from appSettings
  - product and arch are detected when detection the bootloader connection
  - url is calculated based on editor version, product, arch and required firmware from appSettings or CMS

PR #780, opened 2024-07-18
🚩 PR: Enhanced Mode style action blocks and added Endless Mode block

### LINKED ISSUE #772: Add optional min, max field to element mode actionblocks
## Encoder Mode
- [x] Add min max option (only enabled if absolute mode is selected) self:emi(x),self:ema(x)
- [x] Add Sensitivity option (default if 100%) self:ese(x)

## Endless Mode
- [x] New actionblock for Endless Mode based on Encoder Mode
- [x] There is only one Relative mode unlike in the Encoder mode
- [x] short functions used are the following: 
- [x] self:enmo,enmi,enma,env0,ense

## Potmeter Mode
- [x] Add min value field
- [x] Preview teh precalculated resulting step size based on min, max and bit-depth

## Button Mode
- [x] Add min max option self:bmi(x),self:bma(x)
- [x] In step mode preview the resulting step values based on min max and mode

## Common
All of these are optional, when not enabled should not be generated into lua code

elsoazemelet:
Closes #772

- Added additional min/max and sensitivity fields for Button/Potmeter/Encoder mode action blocks
- Added new action block of Endless Mode

**QA request:**

- [ ] Test Endless Mode action block for TEK2 controllers
- [ ] Decide upon the design of the enabled check boxes. Currently the grid-ui kit variant is implemented, which the first figure below.

image


src/renderer/config-blocks/SettingsPotmeter.svelte
```diff
@@ -85,52 +98,88 @@
   ];
 
   let suggestionElement = undefined;
+  let minMaxEnabled = false;
+
+  function calculateStepSize(bit, min, max) {
+    const step = Grid.round((max - min) / (Math.pow(2, bit) - 1));
```
SukuWc, line 104: Correct math here would be (max-min+1)/ 2^bit!

src/renderer/config-blocks/SettingsButton.svelte
```diff
@@ -75,28 +94,100 @@
   ];
 
   let suggestionElement = undefined;
+  let minMaxEnabled = false;
+
+  function calculateStepValues(steps, min, max) {
```
SukuWc, line 99: Example: 3 step mode, min=-10 max=20 displayed steps are incorrectly: 0, 10, 20, 30 (should be -10, 0, 10, 20)

first step should always be the min value, then each step is calculated from that

src/renderer/config-blocks/SettingsButton.svelte
```diff
@@ -75,28 +94,100 @@
...
+        {/each}
+      {:else}
+        <span>N/A</span>
+      {/if}
+    </div>
+  </div>
+  <MeltCheckbox bind:target={minMaxEnabled} title={"Enable Min/Max Value"} />
+  <div class="flex flex-row gap-2">
+    <div class="flex flex-col">
+      <span class="text-sm text-gray-500">Max</span>
```
SukuWc, line 152: That label of the input should be min instead of max!

<span class="text-sm text-gray-500">Min</span>

src/renderer/config-blocks/SettingsButton.svelte
```diff
@@ -75,28 +94,100 @@
...
+    on:validator={(e) => {
+      const data = e.detail;
+      dispatch("validator", data);
+    }}
+  />
+
+  <div class="flex flex-row gap-2">
+    <span class="text-gray-500 text-sm">Step values:</span>
+    <div class="text-white text-sm">
+      {#if Number(bmo) > 1}
```
SukuWc, line 139: if bmo > 0 than steps should be displayed, even in toggle mode!

SukuWc:
## Common problem
- [x] Move suggestions closer to relevant input field group! 
- [x] Move step preview under min max and only preview when min/max is enabled otherwise display message
- [x] If multiple suggestions zones are needed then please follow Led Color block fro reference implementation!
image

## Potmeter mode:
- [x] Calculate step size should be as follows:  (max-min+1)/ 2^bit

## Button mode:
- [x] if bmo > 0 than steps should be displayed, even in toggle mode!
- [x] Rework suggestions as follows: Momentary, Toggle, 3-step, 4-step to correctly indicate step count
- [x] Button min fields label is incorrectly 'Max'
- [x] Min value should be subtracted from calculated step values
Example: 3 step mode, min=-10 max=20 displayed steps are incorrectly: 0, 10, 20, 30 (should be -10, 0, 10, 20)
first step should always be the min value, then each step is calculated from that!
image

Greg-Orca:
related issue #812

PR #782, opened 2024-07-21
🚩 PR: Added new Button Step action block

### LINKED ISSUE #775: Implement button step special actionblock
Special action block that mimics if and elseif blocks to help detect button mode step values using the is_step lua helper function

Each newly added nextStep actionblock (elseif) increments on the index of the step to be detected!

Image

elsoazemelet:
Closes #775 

- [ ] Added button step action block, that works with button mode (2-step, 3-step)
- [ ] Additional steps can be added similiarly to if condition blocks
- [ ] The displayed step name, and the code on grid should update dynamically, when steps are added/removed

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "monaco-editor": "^0.34.0",
     "node-disk-info": "^1.3.0",
     "node-fetch": "^2.6.7",
+    "number-to-words": "^1.2.4",
     "panzoom": "^9.3.0",
     "polka": "^0.5.2",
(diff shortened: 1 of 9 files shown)


SukuWc:
Looks promising, only need couple of minor tweeks:
- [x] Add "Add Action" helper box similar to all other special action blocks
image
- [x] Action picker menu should not recommend nesting of Button Step actions, please remove this option from the list when it it would cause nesting
image
image

PR #783, opened 2024-07-21
🚩 PR: Added block name feature

### LINKED ISSUE #774: Add optional name field for actionblocks
This would used for easier commenting purposes.

- [x] First implement this on the codeblock actionblock
``` --[[@gms#comment for this block]] ``` should be the compiled tag
- [x] Use F2 as shortcut for editing the name if the block is selected
- [x] At the top of monaco modal allow this to be edited ina single text field
- [ ] BLOCKED: Use this name instead of CB #2 name in path display
- [x] Propose UX for editing name of other actionblocks

elsoazemelet:
Closes #774 

- [x] Action blocks can be named when clicked on the edit button on the block header
- [x] CodeBlock action block can be renamed when opened in monaco editor


SukuWc:
Issues found:
- [x] Default config is missing all of the Local variable blocks!
- [x] If user chooses same name as the Default was then it should not be added to Lua to save character count
- [x] Default Code Block name is show in modal after close and re-open
- [x] Adding space to codeblock name (eg: "Test Script") causes is to not be rendered in action list after navigating to different event. It gets parsed into previous action block's last field (see screenshot)
image

SukuWc:
- Clicking Pencil after edit should save
- Hide pencil under dev flag in preferences

PR #796, opened 2024-07-30
Added NRPN MIDI action block

### LINKED ISSUE #734: [RFC] MIDI NRPN Actionblock
https://en.wikipedia.org/wiki/NRPN

Nrpn allows the extension of control parameters beyond CC 0-127
MSB and LSB Controller numbers are sent first to address the control element then MSB and optionally LSB is sent for the control value!

image

This actionblock should send 
- 3 separate gms() calls for (14bit address, 7bit value) or 
- 4 separate gms() calls for (14bit address, 14bit value) controls

No firmware modification required as far as I see!

What do you think? Is this widely supported by hardware and DAW's?

elsoazemelet:
Closes #734


narayb:
From a usability standpoint fields for both MSB and LSB parameters are desireable instead of the singular field we have now.

kép

It is also how manufacturers describe NRPN messages received on their devices in their user manuals:

kép

narayb:
one bug remains:
1, input discrete values into the LSB MSB fields or into the NRPN CC field
2, commit the changes by clicking outside 
3, switch to another event or in some other way re-render this block
kép
4, '?' will always show up in the NRPN CC field instead of the actual calculated value

narayb:
When writing number into NRPN CC field on a Mac machine, unintended parentheses are rendered. This causes the tests to fail.

kép

Greg-Orca:
NRPN is made for high resolution (14-bit), so it would be more convenient if 14-bit was the default.
Screenshot 2024-08-27 at 10 58 40

Greg-Orca:
The parentheses seem unnecessary in both cases.

Screenshot 2024-08-27 at 10 38 40
Screenshot 2024-08-27 at 10 38 54

Issue #754, opened 2024-06-18
access_tree has multiple references in action blocks, but doesn't seem to be used, could be deleted


danim1130:

SukuWc:
Access tree will be removed one updated runtime store implementation is in place!

PR #820, opened 2024-09-11
Refactored runtime, added path display for CodeBlock

### LINKED ISSUE #811: Runtime refactor: User Array of Action stores instead Action list string

elsoazemelet:
Closes #811, #821, #760, #754, #820, 

Reworked runtime (Affected parts of editor must work as before this change set):

- All toolbar/context menu actions are affected (copy, paste, etc.).
- Disabled states are affected for the toolbar
- Context menu states may differ from toolbar states
- Code Block and Monaco editor is reworked 
- Character limit calculation is reworked, in Monaco editor it is differently calculated than in DebugView
- Action block naming is reworked, Monaco editor name editing should match up with Action block name editing. Monaco editor and the action block name should always match, no matter which was edited, and where.
- Preset/Profile load is affected.
- Added code path for Monaco editor, that should alway correctly display the position of an Action block
- Everything above should also be working in Virtual Mode
- Other parts of editor may be affected that can be in connection with a displaying/modifying scripts on a grid module.


narayb:
- [x] After commiting changes in code block the path ending changes to say "Code Block" instead of "Block 2".

incorrect_name

elsoazemelet:
- [x] No configuration panel reload

Greg-Orca:
The tests are failing because the comment blocks are broken and not saving the values.

SukuWc:
- [x] FI from stable
- [x] Add profile throws exception
- [x] Suggested element not working in profile cloud
- [x] Load preset/profile broken: Cannot read property (parent) of undefined method: getInfo()
- [x] Config is not fetched automatically after clear and discard all
- [x] Move proflie load logic from Overlay.svelte to operations.ts

Greg-Orca:
- [x] BUG: The add action is **not working on empty** element, only the bottom "Add action blocks" works.
Screenshot 2024-10-10 at 11 12 31

Greg-Orca:
- [x] BUG: Copy Element then Overwrite Element deletes all actions!



https://github.com/user-attachments/assets/fe93a1d0-27ee-4e73-af2f-8da2e11f35da

PR #822, opened 2024-09-16
Added character limit display for editor and restrictions for character input

### LINKED ISSUE #801: Implementing Character Limit and Overwrite Prevention in Action Block Fields
The goal of this implementation is to ensure that users are aware of the character limits in action block fields and to prevent the loss of progress due to character limit exceedance. This will be achieved by blocking overwriting when the character limit is reached and providing a visual counter to assist users in managing their text input.

## Current Issues
**1. Character Limit Overwrite Issue:**
When users reach the character limit in an action block field, any further input is not saved, resulting in the loss of unsaved text.

**2. Lack of Character Feedback:**
Users currently do not have a clear indication of the remaining characters available in the action block fields, leading to potential frustration and loss of work.

## Implementation Plan:

1. Integrate a detection mechanism within each action block field to monitor the number of characters entered in real-time.
2. Develop functionality to block additional input once the character limit is reached. This prevents any further characters from being added, ensuring users do not lose unsaved text.
3. Add a dynamic counter at the top of Element that updates as users type, showing the count.

Screenshot 2024-08-07 at 13 46 36

elsoazemelet:
Closes: #801


Greg-Orca:
The tests are failing because the comment blocks are broken and not saving the values.

elsoazemelet:
- [x] LineEditor can break char limit
- [x] Local/Self/Global check (no commit available)
- [x] Visual bug of char limit being over list of action blocks

elsoazemelet:
- [ ] Button mode min/max does not dispatch?

PR #831, opened 2024-10-02
Add social login support for web

### LINKED ISSUE #743: Profile Cloud Register and Social Login does not work on web

danim1130:


Greg-Orca:
It would be nice if the login worked by pressing 'Enter'. Related: https://github.com/intechstudio/profile-cloud/issues/92

It would be more aesthetic to have a spinning indicator while the system fetches data.
https://github.com/user-attachments/assets/5b608c71-96a2-4761-b632-4ad409ed560d

Greg-Orca:
**User enumeration** risk in our password reset flow, as the system provides explicit feedback indicating whether an email is not registered.
Screenshot 2024-10-04 at 13 32 49

Greg-Orca:
After logging in, reloading the page incorrectly displays the user as not logged in, though they remain authenticated in the background. This bug occurs because, upon refresh, the Cloud settings jump back from Nightly to Prod, causing login and logout functions to fail.

https://github.com/user-attachments/assets/0afdf67d-4ca7-46d5-ae67-077f7a21f55d

Greg-Orca:
It would be more aesthetic to have a spinning indicator while the system fetches data.
https://github.com/user-attachments/assets/5b608c71-96a2-4761-b632-4ad409ed560d

Greg-Orca:
The "Can't register" message is no longer valid on Nightly, as Google Sign-In also functions as a registration method when using a new email address. Additionally, if a password reset is requested for an email used only for Google Sign-In, the user can set a new password and log in without using Google Sign-In.

Screenshot 2024-10-04 at 14 14 53

PR #849, opened 2024-11-04
Reworked Dragn-N-Drop, added MultiEvent edit for developer options

### LINKED ISSUE #483: Turbo multi-lane configuration of control elements with drag and drop
If the configuration panel is narrow then normal view, if wide then the proposed idea:

- [ ] Integrate multi event interface into configuration sidebar
- [ ] If developer option is enabled, then sidebar should show all events simultaneously for the selected element
- [ ] Selecting actionblocks must be limited to a single event
- [ ] Paste action button should be disabled when using multi-event, each event needs separate paste button
- [ ] Implement responsive design by automatically switching between single and multi-event views

Image

elsoazemelet:
Closes #483


elsoazemelet:
- [x] Cross-Dragging actions to first position visually appears disabled

elsoazemelet:
- [x] Drop the border concept, integrate separator if necessare instead
- [x] Bring back helper add/paste
- [x] In single lane, bring back paste icon
- [x] Bring back export button only in single lane view
- [x] Reactive view change should be based on percentage value instead of pixel value
- [ ] Fix tests, fix not working base functions (Add action between two action blocks, drag and drop inside if)
- [x] Multi lane add/paste should be horizontal
- [x] Single lane paste button should not be visible

PR #862, opened 2024-11-25
MeltCombo is used from grid-uikit

### LINKED ISSUE #860: Combo Cox to UI-kit

elsoazemelet:
Closes #860


elsoazemelet:
- [x] Melt combo does not register input
- [x] Change context menu store value to undefined instead of using null
- [x] Revert element selection visibility detection to original

PR #902, opened 2025-01-12
Chore: Apple notarization flow

kkerti:
~~1. Googled a lot about `Unexpexted error ... JSON...` and found similar issues, all related to authentication issues.~~
~~2. Bumped `electron-builder` to the last version in `24` major. This was needed to confirm the issue more easily. We should update both electron, node and electron-builder soon.~~
~~3. Removed `electron/notarize` package as we do not use separate sign script, but electron-builder directly.~~
~~4. Issue was that the `APPLE_APP_SPECIFIC_PASSWORD` in github secrets was invalid. This password is automatically revoked when the related apple_id password is reset. **I could not login to my apple account due to some issue with apple itself, so had to recreate my password**. After securing my own account, had to re-create the app specific password.~~
~~5. An other issue seemed to me that the APPLE_TEAM_ID was invalid, but could not confirm it. Updated it anyway.~~

~~Basically now all recently failed runs can be re-run.~~

Sporadic issue, impacting other projects as well. Based on recent suggestions from others, added a retry script.

- Thread about possible solutions. https://github.com/electron/notarize/issues/219
- Inspiration: https://github.com/KittyCAD/modeling-app/issues/4199
- Possbile solution to automate resolution: https://github.com/KittyCAD/modeling-app/pull/4892/files

### DIFF
--- a/.github/workflows/build-matrix.yml
+++ b/.github/workflows/build-matrix.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-13]
+        os: [ubuntu-latest, windows-latest, macos-13]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


SukuWc:
@danim1130 please review and merge if all good

PR #905, opened 2025-01-15
Reworked Local Variables action block

elsoazemelet:


SukuWc:
- [x] FI from main to fix macos build
- [x] Drag-and-drop changing the order of actionblocks does not trigger update to be sent to grid!

PR #907, opened 2025-01-20
Fixed analytics bombardment on update action

### LINKED ISSUE #906: Change analytics on update action to sync action
- Analytics data fields should remain the same

elsoazemelet:
Closes #906

### DIFF
--- a/src/renderer/runtime/operations.ts
+++ b/src/renderer/runtime/operations.ts
@@ -207,32 +207,15 @@ export async function syncWithGrid(target: GridAction) {
 export async function updateAction(
   target: GridAction,
   data: ActionData,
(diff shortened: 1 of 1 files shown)


src/renderer/runtime/operations.ts
```diff
@@ -205,6 +218,9 @@ export async function updateAction(
     })
     .catch(handleError)
     .finally(() => {
+      if (!syncWithGrid) {
```
SukuWc: syncWithGrid parameter should call syncWithGrid instead of sendToGrid which already contains the analytics code. This guard clause and the following analytics track call are redundant, i think.

PR #927, opened 2025-01-29
Fix display of midi channel

claytron:
Unsure if fixing this in the template layer is the proper fix, but saw that being done elsewhere.

Was seeing the 0 based raw MIDI channel in the monitor.

Screenshot 2025-01-28 at 18 37 26
Screenshot 2025-01-28 at 18 37 36

But the actual value was this:

Screenshot 2025-01-28 at 18 37 51


SukuWc:
This looks good, thanks for the contribution!

PR #928, opened 2025-01-29
Added !NEW! SimpleColor action block

### LINKED ISSUE #912: Implement new color action block based on reference design
Use current 2D colorpicker insted of the proposed color cycle for now!
Use vertical alpha slider instead of the proposed intensity rotary dial for now!
 '+' (Add keyframe) button should not be on the preview display, it should be in the top next to the name of the actionblock)
'-' (Remove keyframe) button is position next to it

Clicking the '+' button adds second (min) key frame. (new '+' button appears in the middle of the range)
Clicking the '+' button adds third (mid) key frame (maximum number of key frames is 3)

Each active key frame can be selected on the preview display, R, G, B and alpha values are loaded for the given key frame.

Api specification
```
-- Using single color parameter:
self:led_color(layer, {{r,g,b,a}})
-- Using min and max keyframes:
self:led_color(layer, {{r,g,b,a},{r,g,b,a}})
-- Using min, mid and max keyframes:
self:led_color(layer, {{r,g,b,a},{r,g,b,a},{r,g,b,a}})

```
Notes:
Add combobox for element selection:
- Default: 'Self' (self:element_index())
- Element 'Element 0' (element[0])
- _element_: 'Set manually by the user'
Add combobox for layer selection:
- _self:led_layer()_ 'Automatic' (Default!)
- _1_ 'Button/Potentiometer/Fader'
- _2_ 'Encoder/Endless rotation'

Questions: 
- What happenes with beautify?
  - Handled in firmware 
  - Beautify is added in lua microcode of the led_color method defined as member function
- Is alpha parameter 0...255 integer or 0...1 floating point?
  - 0...1 floating point represented with 2 decimal places
  - r, g, b parameters are still 0...255 integer
  - If alpha parameter is omitted then it will be considered 1 by lua microcode
- How does the actionblock determine element number?
  - Uses self to determine it in firmware during runtime by default!
- How does the actionblock determine layer number? 
  - Based on self:led_layer() in runtime
  - Based on formula: layer = 1+eventtype%2

Image
Image
Image
Image

elsoazemelet:
The expected behaviour and design proposal is described in the ticket that this PR closes. See below.

Closes #912


danim1130:
Can we move the ColorPicker components to the uikit library?

Greg-Orca:
### Test Failed:
**Issues:**
- Module LEDs do not receive the color.
- Not working: Each active keyframe can be selected on the preview display, but R, G, B, and alpha values are not loaded for the given keyframe.
- Cannot write RGB and alpha values in the text box; displays 'NaN'.

env: MacOS

https://github.com/user-attachments/assets/a6c0457f-12fb-4de4-b745-3b9113f89746

narayb:
We'll issue an updated design for the colorbar as it looks too much like a slider and causes the user expectation that it's clickable and slide-able.  That's on the design, we'll fix it.

Aside from this the feature looks as specified. Nice!

However I can only echo the sentiments of @Greg-Orca above:

- [x] slider does not update position or values when switching between color states min, mid or max
- [x] layer 0 does not exist, therefore it should not be a default layer (layers are 1 or 2)
- [x] typing into the value fields does not produce a change, fields become `NaN`
- [x] any kind of configuration created doesn't run on the nightly firmware currently, possibly awaiting a new FW patch to implement the functionality itself (this could be the cause of the above problems as well?)
- [x] the parameters should match already existing, similar blocks in design (LED Number input field followed by Layer input field)  

kép


- [x] resizing the panel causes the colorwheel to grow unnecessarily large, the colorwheel size should probably stay static

color1

- [x] 'Easy Color' is not the name I would give to the block itself, something akin to `Simple Color` would suit it better
- [x] the randomization of a new added color wasn't part of the specification, and the way Editor hadles it causes it to change the configuration constantly

color2

Also the sequence of events when adding new colors is wrong. Now what it does is:
1, Has a color at the max, then you add a second color.
2, The color previously at the max moves to the min position.

What it should be doing instead:
1, Has a color at the max, then you add a second color.
2, The color previously at the max should **still stay at the max position and the second color should move into the min position.**

elsoazemelet:
- [x] Color picker Square Y axis and Circle radius should set brightness 50 to 100 and render picker accordingly
- [x] Selection order when adding new should follow visual adding (max, min, middle)
- [x] Selection should be more visible (not pretty)
- [x] Fix square pickers out of bound selection
- [x] Make VM more lightweight

Greg-Orca:
### Test Failed
- Variables are not working in the RGBA input fields
Screenshot 2025-02-19 at 13 50 24

- HSL sliders jumping while adjusting



- Still cant write any parameters to the RGBA input fields

Greg-Orca:
### Test Failed
 - Variables not working in RGBA, Classic, and HSL input fields
 - The circle is out of the color picker when using a low number or 0
Screenshot 2025-02-20 at 14 10 51
Screenshot 2025-02-20 at 14 12 11

 - circle is not jumps to the corresponding RGBA and Classic value
Screenshot 2025-02-20 at 14 25 46

 - HSL still jumps while adjusting parameters

Screenshot 2025-02-20 at 14 1

narayb:
## Color picker design change

The current design fails to serve the user on the following points:
- it has no way to display dark colours on the colour picker
- the user is not able to pick dark colours on the colour picker
- the intensity does not fulfill the part of "darkening" either

The Solution is to use a design that solves these issues.
simplecolorpicker
*Taken from https://rgbcolorpicker.com/*

The truth is, for this the block has to work like this:

1, It has fields for color code entry in multiple formats (RGB, HSV, HSL, HEX etc), but has only one color picker to display the result.
2, Color code entry happens and then gets transformed to a format that can be properly displayed on the - normally - HSL color picker above.

Greg-Orca:
### Test Failed
- Color change not synchronized while dragging and dropping the picker; only when releasing the picker. (The previous version had this functionality.)
Screenshot 2025-02-21 at 12 59 16


- at RGB 0, 0, 0 picker out of the circle
Screenshot 2025-02-21 at 12 59 16

 - The picker is not perfectly aligned with the color palette. On the right side, I can pull it out, but on the left side, I cannot.
Screenshot 2025-02-21 at 12 59 16

 - Opening and closing the Action block fails to maintain the chosen color picker state.
 Screenshot 2025-02-21 at 12 59 16

elsoazemelet:
- [ ] Preview should work
- [x] Persistent color selected (example: $appSettings.persistent.keyboardLayout)
- [x] Move this under feature flag

Greg-Orca:
### Test Failed
- Classic and HSL color pickers are lagging.
- Still no real-time color visibility on the slider like before.

ezgif-89aece9d949061

elsoazemelet:
> ### Test Failed
> * Classic and HSL color pickers are lagging.
> * Still no real-time color visibility on the slider like before.
> 
> ezgif-89aece9d949061 [ ezgif-89aece9d949061 ](https://private-user-images.githubusercontent.com/88966111/419045275-8402c1e2-7c6f-42db-b1c2-f320fc5079fb.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDExODUyMTgsIm5iZiI6MTc0MTE4NDkxOCwicGF0aCI6Ii84ODk2NjExMS80MTkwNDUyNzUtODQwMmMxZTItN2M2Zi00MmRiLWIxYzItZjMyMGZjNTA3OWZiLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTAzMDUlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUwMzA1VDE0MjgzOFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTViZDgxMGRiYmU4YzY1MTcwNGMxZjY4NmM3OWYwZmM2MTAzM2RlYmNjMGUzZjhkNDZjZThjNTYwNzE2YTY4ZGYmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.sj8PeagBnEJArLJhTpeVcfNZFLp3cmmfvqH1BwyILbA) [ ](https://private-user-images.githubusercontent.com/88966111/419045275-8402c1e2-7c6f-42db-b1c2-f320fc5079fb.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDExODUyMTgsIm5iZiI6MTc0MTE4NDkxOCwicGF0aCI6Ii84ODk2NjExMS80MTkwNDUyNzUtODQwMmMxZTItN2M2Zi00MmRiLWIxYzItZjMyMGZjNTA3OWZiLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTAzMDUlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUwMzA1VDE0MjgzOFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTViZDgxMGRiYmU4YzY1MTcwNGMxZjY4NmM3OWYwZmM2MTAzM2RlYmNjMGUzZjhkNDZjZThjNTYwNzE2YTY4ZGYmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.sj8PeagBnEJArLJhTpeVcfNZFLp3cmmfvqH1BwyILbA)

- Rendering preview color as a feature is halted currently. Out of scope.
- CSS lagging or rendering optimalization is (sadly) also out of scope.

Greg-Orca:
>Rendering preview color as a feature is halted currently. Out of scope.
>CSS lagging or rendering optimalization is (sadly) also out of scope.

**Both worked in this branch perfectly in the beginning. Out of scope?**

SukuWc:
1. Lagging is only present on macos, @danim1130 will investigate
2. Please re-enable the live previewing of color that was present in previous commit [262c4dd](https://github.com/intechstudio/grid-editor/pull/928/commits/262c4dd53edd236244488eff338561a3c38ff7ea)

elsoazemelet:
The transformation between HSL (used in our color pickers) and RGB (used in the code) during value binding (live preview) results in some "jitter" in certain, if not all, pickers when selecting desaturated or darker colors. However, this issue does not arise with the square and circle pickers, as they are unable to select desaturated or dark colors.

That being said, as reported, the HSL slider picker is affected by this issue. At present, I am unaware of a solution currently that would address this without introducing unnecessary code complexity.

Additionally, following our discussion two weeks ago, the live preview issue has led to this feature being classified as a developer feature.

To my knowledge, this iteration did not include a resolution for this particular issue, which is why I referred to it as being out of scope.

Further investigation into resolving the live preview issue is ongoing, but it may fall outside the scope of this iteration.

elsoazemelet:
Discussed the above mentioned situation of live preview with @SukuWc  in person. Live preview is implemented back to solution.

Greg-Orca:
### Test Result
**Passed:** 
 - smooth transitions
 - live update
 - picker is always in correct positions

**Failed:**
- Intensity not updates now if I switch in color "stage" 
ezgif-40764767478105

- Intensity can't be adjusted only to the middle stage; it influences the right endpoint
ezgif-449ca62b419562

SukuWc:
Functionality can now be tested on real modules!
1. Get a real module
2. Update to nightly firmware: https://github.com/intechstudio/grid-fw/raw/preview/Preview/Firmware/grid_esp32_nightly.uf2
3. Clear all configuration
4. Select the system element and add the following code to its setup event:
``` 
for i = 0, #element do
    element[i].led_color = function(self, l, c)
        print("Easy color triggered", l, #c)
        local up = table.unpack
        local i, x, y, z = self:element_index(), c[1], c[2], c[#c]
        if #c == 1 then
            x = {0, 0, 0, 0}
            y = {z[1] // 2, z[2] // 2, z[3] // 2, z[4] / 2}
        elseif #c == 2 then
            y = {(x[1] + z[1]) // 2, (x[2] + z[2]) // 2, (x[3] + z[3]) // 2, (x[4] + z[4]) / 2}
        elseif #c == 3 then
        end
        led_color_min(i, l, up(x))
        led_color_mid(i, l, up(y))
        led_color_max(i, l, up(z))
    end
end
print("Easy color installed")
```
5. Now you can use the Easy Color action on the control elements.
6. Note that led intensity should be set to -1, this causes the firmware to calculate the correct color mix between minimum and maximum value.

Screenshot_2025-03-16_00-32-53

SukuWc:
Bug found:
Adding or removing keyframes send Syncs with grid before sending new config out (reactivity order is reversed, dispatch("sync") runs befor the reactive $: statment on the data store)

```
  function handleAddLayer() {
    const last = get(data).previewColors.at(-1);
    data.addLayer(last);

    dispatch("sync");
  }

  function handleRemoveLayer() {
    const selected = get(data).selectedIndex;
    data.removeLayer(selected);

    dispatch("sync");
  }
```

Greg-Orca:
### Test Failed
The editor freezes when I use a function (e.g., self:encoder_value()) in an input field and then close and reopen the Simple Color block (or copy and paste)

Screenshot_2025-03-18_at_15 38 58

SukuWc:
Parsing bug fixed for red, green, blue and alpha input fields!
Num and layer fields are special due to underlying parsing strategy, these do not produce runtime error but parsing can be incorrect!

Separate issues created to track upcoming work: 
https://github.com/intechstudio/grid-editor/issues/1040
 https://github.com/intechstudio/grid-editor/issues/1041

Issue #922, opened 2025-01-25
Adding a new key-value pair to self action block crashes Editor, making it unresponsive


kkerti:
Issue:
When **commit** is used on the _self_ action block and character number is almost 400, Editor crashes.

Reproduction steps:
1. Add a characters to config, so it is almost 400 or try out the _Mouse takeover vst controller 1.1_ preset's self block on the setup event - that's almost at 400 characters.
2. Add a _self_ action block or fill the action block with new entries
3. On commit, Editor crashes.

![Image](https://github.com/user-attachments/assets/5bc648c6-b142-422c-b0c4-f775be442e9e)

kkerti:
The issue is also true to _local_ action block, probably it will crash on _global_ too.

SukuWc:
We will fix this along with #914

Issue #904, opened 2025-01-14
14-bit MIDI Blocks with Lua Math Library Deletes Input Field


Greg-Orca:
When attempting to copy and paste or add a new 14-bit MIDI blocks when using the math library, the input field gets unexpectedly deleted/ cleared. 

Editor 1.3.8 
MacOS 15.1.1 

https://github.com/user-attachments/assets/9fc66ee9-ad89-4063-aef0-06016d6fee80

SukuWc:
Use lua syntax checker on every combobox, remove any custom code that uses regex to count commas, parenthesis etc.

elsoazemelet:
- [ ] intech_lua.json

PR #929, opened 2025-01-29
Deprecated: Various Changes

### LINKED ISSUE #914: Combine the local, self and global actionblocks into a single svelte component. The three different variants should still be avalilable from the action picker!

elsoazemelet:
Closes #914 , #922, #904

### **Key  changes and added features:**

- Syntax error handling is reworked for all blocks (see working below)
- Logic of runtime is changed to accomodate new syntax error handling
- Editor operation states are now affected by syntax error (see working below)
- Changed LineEditor component (see changes below)
- Variable manager component is added to Self/Global/Locale action blocks to unify the logic and esign of these blocks

### **New syntax error handling:**

- Individual action blocks do not handle syntax error handling (affected blocks: variable type blocks).
- ActionBlocks with syntax error is no longer sent to grid
- The last syntax correct configuration of a block is called now synced. The synced state is updated when a block is succesfully sent to grid. This state has to be always syntactically correct
- Blocks that are affected with syntax error are marked with red outer border for the whole block
- Blocks with syntax error are reverted to synced state as soon as they are out of scope (e.g.: changing events)
- Blocks values are updated when input fields lose focus (this logic is unchanged)
- ComboBox type input fields still have validators (red outline on validation error), and these are only checking if the inpuit value is empty (this logic is unchanged)
- Out of scope action blocks actual and synced states **MUST** always be consistent. Meaning: syntax correct action blocks must be always sent to grid, and its value can not change even when reloading the app, or changing panels, events, etc.

### **New operation states:**

Operations that affect targets with syntax error are disabled
**Example:** selecting and trying to copy a block or group of blocks that has syntax error results in disabled copy action toolbar button

**Affected operations:**

- Cut Actions
- Copy Actions
- Merge Actions
- Copy Element
- Store configuration

### **LineEditor changes**

- LineEditor is no longer dependent on the ActionBlock, it uses runtime independent parameters instead to track available character limits and suggestion scope narrowing.
- The LineEditor should work as before regarding character limit handling
- Suggestions of the LineEditor during code typing must be consistent with previous releases


elsoazemelet:
- [x] Remove type dependent stuff from shared component and move it into the wrapper
- [x] Remove type specific label from add, use generic text
- [x] Make preprocessor functions safe (local should be trimmed instead of replaced, self. replacing should be only on the leftside of assignment
- [x] Fix tests
- [x] Fix #922 here
- [x] Remove unintended commits

elsoazemelet:
- [x] Fill out PR to be QA ready
- [x] Disable syntax dependent operations when error is present
- [x] Note key feature changes and new use-cases for QA in this PR
- [x] Expose syntaxError in DynamicWrapper and use some counting method to determine global state (onDestroy?)
- [x] Add scope parameter to line editor to determine element type
- [x] Remove parenthesis counting and unnecesarry crafted error detection

narayb:
### Test failed

Did not even test all changed features.

NRPN solution created new issue:

colon_bug

Adding new local var pair freezes Editor, deletes all previous:

local_bug

Config panel autoscrolls when hovering over top and bottom area:

autoscroll_bug

narayb:
Ignoring the scrolling issue and the NRPN issue, testing out the Local Block, a different kind of issue appeared:

local_bug_disappearing_variable

The previously encountered issue (local block becoming empty when adding new pair) is now fixed.

Suggesting splitting the issue, remove everything that has nothing to do with the variable Action Blocks @elsoazemelet @SukuWc.

PR #943, opened 2025-02-02
Added multi event view to developer settings

### LINKED ISSUE #931: RFC: Create accessible design for multi-event view
Image

This should be a checkbox. And it should be off by default.

elsoazemelet:
Closes #931 

- Multi-even view is disabled by default until proper design is ready for implementation!
- Removed MultiEvent view dialog from developer settings
- Added description for MultiEvent view and button to enable/disable it in developement settings


Greg-Orca:
Bug:
The Editor freezes when selecting all elements, dragging one to another column, and using the merge or delete function.
https://github.com/user-attachments/assets/41f712bf-1fd3-4a30-84ec-6ec2eb91f219

Greg-Orca:
### Test Passed
 Multi-Event View is disabled by default and can be enabled/disabled in Developer Settings and correct description.

PR #984, opened 2025-02-20
Merged variable blocks into using a single shared component

### LINKED ISSUE #914: Combine the local, self and global actionblocks into a single svelte component. The three different variants should still be avalilable from the action picker!

elsoazemelet:
Closes #914 

- Changed LineEditor component (see changes below)
- Variable manager component is added to Self/Global/Locale action blocks to unify the logic and design of these blocks

### **LineEditor changes**

- LineEditor is no longer dependent on the ActionBlock, it uses runtime independent parameters instead to track available character limits and suggestion scope narrowing.
- The LineEditor should work as before regarding character limit handling
- Suggestions of the LineEditor during code typing must be consistent with previous releases


Greg-Orca:
### Test Failed
Variable blocks are not working; the module is not receiving the correct values.
ezgif-29ec9ffdc19cfd

Greg-Orca:
### Test failed
Incorrect value freeze Editor.

Greg-Orca:
### Test failed
Incorrect value in the variable field, freezes editor.
Screenshot 2025-03-03 at 16 25 17

Greg-Orca:
### Test Failed
- **Global** immediately freeze editor with incorrect value
- Closing the **Local and Self** block with Incorrect value in the variable field, freezes editor.
ezgif-65b4d321a71713

Greg-Orca:
Passed: above issued fixed
Failed: When adding a new pair and typing into the variable name field, the focus shifts away after the first character is typed.
ezgif-6c26a0b8338b44

Greg-Orca:
### Test Failed
Short truncated names jump to the original name, (no suggestion list). Tested on another version, this bug is present only this PR.
ezgif-8b04c9a384f153

Greg-Orca:
### Test Passed

PR #985, opened 2025-02-20
Updated syntax check and error handling logic

### LINKED ISSUE #982: bug: Mode Action field clear freezes UI
**Describe the bug**
Clearing any of the fields inside the mode action causes an exception to be thrown and the UI to freeze, only UI restart fixes it.

**To Reproduce**

1. Add a Mode action to a module
2. Clear any of the Combo fields with backspace

**Expected behavior**
Error state can remain, but application remains operable.

- OS: Windows 11 (error seems platform independent)
- Editor version: Latest nightly ( commit 8907eac59b7e0f88e2987546d0d4a8a64858019e )

**Additional context**
Tested with all mode actions, but not every field

Image

elsoazemelet:
Closes, #922, #904

### **Key  changes and added features:**

- Syntax error handling is reworked for all blocks (see working below)
- Logic of runtime is changed to accomodate new syntax error handling
- Editor operation states are now affected by syntax error (see working below)

### **New syntax error handling:**

- Individual action blocks do not handle syntax error handling (affected blocks: variable type blocks).
- ActionBlocks with syntax error is no longer sent to grid
- The last syntax correct configuration of a block is called now synced. The synced state is updated when a block is succesfully sent to grid. This state has to be always syntactically correct
- Blocks that are affected with syntax error are marked with red outer border for the whole block
- Blocks with syntax error are reverted to synced state as soon as they are out of scope (e.g.: changing events)
- Blocks values are updated when input fields lose focus (this logic is unchanged)
- ComboBox type input fields still have validators (red outline on validation error), and these are only checking if the inpuit value is empty (this logic is unchanged)
- Out of scope action blocks actual and synced states **MUST** always be consistent. Meaning: syntax correct action blocks must be always sent to grid, and its value can not change even when reloading the app, or changing panels, events, etc.

### **New operation states:**

Operations that affect targets with syntax error are disabled
**Example:** selecting and trying to copy a block or group of blocks that has syntax error results in disabled copy action toolbar button

**Affected operations:**

- Cut Actions
- Copy Actions
- Merge Actions
- Copy Element
- Store configuration


Greg-Orca:
### Test Failed
- Operations are not disabled when cause a syntax error while selecetd
- Operations are not enabled if the syntax error is corrected while is selected.
ezgif-7fda935e8e334c


- After a syntax error, the next input field is deleted. repro: type 1.3,2 _result: 1.3_
ezgif-71fee80fee705f

elsoazemelet:
- [x] check toolbar enabling
- [x] Try using a syntaxpreprocessor for value (a=(any vlaue))

Greg-Orca:
### Test passed
**Note:** If one of the action blocks has an incorrect input field and you switch to another element, the input value jumps back to the last automatically saved correct one.

src/renderer/main/panels/configuration/components/Toolbar.ts
SukuWc, line 1: invalid is stronger than checkSyntax!!

src/renderer/main/user-interface/ActiveChanges.svelte
SukuWc, line 1: use invalid instead of checkSyntax

SukuWc (commented):
easy fixes coming soon

PR #991, opened 2025-02-24
Fix package action handling when loading/unloading package

### LINKED ISSUE #918: Missing action block code update converts to default code block
- [ ] Make RAW code selectable

### LINKED ISSUE #919: Select all can select unselectable actions (RAW code)
Fix selector functionality

danim1130:
Change how Editor handles cases when a package gets loaded/unloaded that has an action element in the current configuration panel.
The aim of this PR is the following:
- Automatically change the action window to the loaded/unloaded action state when the corresponding package is loaded/unloaded.
- If an unloaded action is shown, pressing the update button converts it to the correct code block representation.
- Be able to copy, move unloaded action block


Greg-Orca:
### Test Failed

Spotify Package deselect automatically after using the 'Spotify-Like' action black
Screenshot 2025-02-27 at 14 03 39

Char limit indicated incorrectly when merge
Screenshot 2025-02-27 at 14 03 39

Photoshop Menu shortcuts action block throws error
Screenshot 2025-02-27 at 14 03 39

danim1130:
# Test results:

## Spotify bug

### Reason
This is caused by the user not being logged in, and the like action executing, causing a spotify web request exception, which automatically disables the package.

### Proposed solution
The exception will be caught and gracefully handled in the spotify package. No work to do on this PR.

### Further work
Why is action executed when the package is enabled? Is this the default behaviour of actions, or specific to packages? What should be the default behaviour?

## Char limit problem

### Reason
The difference in character count is caused by the short code of the action:
- Photoshop action: `--[[@xpsqa]] gps("package-photoshop", "quick-action", "switch-colors")`
- Merged as code: `--[[@cb]] gps("package-photoshop", "quick-action", "switch-colors")`
Note that the same happens with built in actions where the short code is not two characters, such as the Intensity action.

### Proposed solution
Short codes are neccessary to correctly identify the actions. No further progress can be made within the current architecture.

## Photoshop Menu shortcuts action block throws error

### Reason
The shortcut action needs the suggestions data from photoshop. If photoshop is not running, it tries to read from an undefined variable.

### Proposed solution
Error will be handled in the photoshop package. No work to do on this PR.

PR #1004, opened 2025-03-05
Fixed headers, and for loop header

### LINKED ISSUE #964: Loop-Block Visual and Field bug
### Issue
The Loop block's extra field extends beyond the block background, affecting visual consistency. Additionally, the Repeat field has multiple input-related issues:

- The field loses focus after entering the first digit, requiring a second click to continue typing.
- Pressing backspace inserts an unintended "2" instead of just deleting the character.
- Entering a non-numeric character when the field is empty results in a question mark ("?") appearing.
### Steps to Reproduce:
1. Click on the Repeat field to enter a number.
2. Start typing a number (e.g., "12").
3. Press backspace to delete the number.
4. Try entering a non-numeric character (e.g., "a", "b", "c").
### Expected Behavior:
- The Repeat field should stay in focus while entering numbers.
- Backspace should only delete characters without inserting unwanted values.
- Entering non-numeric characters should't input a "?"

### Env:
MacOS Sequoia
Windows
Editor 1.5.0

https://github.com/user-attachments/assets/8962bd6b-8489-4b43-9ce7-d03c080f64d1

elsoazemelet:
Bugfix. Bug description can be found in original ticket.

Related to #964 (does not close it because it contains multiple issues)

### DIFF
--- a/src/renderer/config-blocks/headers/ButtonStepFace.svelte
+++ b/src/renderer/config-blocks/headers/ButtonStepFace.svelte
@@ -55,7 +55,7 @@
 <!-- svelte-ignore a11y-click-events-have-key-events -->
 <!-- svelte-ignore a11y-no-static-element-interactions -->
 <div
-  class="px-2 w-full h-full {config.information.rounding === 'top'
+  class="px-2 w-full {config.information.rounding === 'top'
     ? 'rounded-tr-xl'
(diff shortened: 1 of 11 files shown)


Greg-Orca:
### Test Passed
Headers are displayed correctly

PR #1007, opened 2025-03-05
Fixed monaco suggestions

### LINKED ISSUE #934: Monaco suggestions duplications should be fixed
- Suggestion duplications are present
- Suggestions does not seem to work (no added self: or element[0])
- Suggestion creation is confusing in code and causes above issues

elsoazemelet:
Bugfix. Bug description can be found in original ticket.

Key fatures fixed:

- Monaco should display correct adressing of element specific function suggestions (e.g.: self:potmeter_value on elements, and element[0]:potmeter_value)
- Monaco should display each suggestion once

Closes #934

### DIFF
--- a/playwright-tests/tests/blocksTests.spec.js
+++ b/playwright-tests/tests/blocksTests.spec.js
@@ -309,3 +309,30 @@ test.describe("Input field keyboard shortcuts", () => {
     await expect(actualValue).toBe(expectedValue);
   });
 });
+
+test.describe("Monaco Sugestion", () => {
+  test.beforeEach(async ({ page }) => {
+    connectModulePage = new ConnectModulePage(page);
+    modulePage = new ModulePage(page);
+    configPage = new ConfigPage(page);
+    keyboardActions = new KeyboardActions(page);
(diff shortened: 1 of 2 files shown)


Greg-Orca:
### Test failed
After `element[x]` , the suggestion still includes `self:`, but it's a syntax error. It should suggest without `self:` in this case
Screenshot 2025-03-06 at 14 52 10

elsoazemelet:
> ### Test failed
> After `element[x]` , the suggestion still includes `self:`, but it's a syntax error. It should suggest without `self:` in this case Screenshot 2025-03-06 at 14 52 10

Out of scope of this ticket, possibly intended working? Needs discussin with @SukuWc

Greg-Orca:
Discussed with @SukuWc. It should be displayed without `self:`

Greg-Orca:
Syntax highlighting is different. The functions were yellow before, but now only the global functions are yellow.

Screenshot 2025-03-12 at 13 03 59

Greg-Orca:
functions are not highlighted in yellow color

Greg-Orca:
### Test Passed

PR #1044, opened 2025-03-19
SUKU fixed line editor suggestion scopes

SukuWc:
Fixed suggestion for line editor components in the following action blocks:
- Local
- Global
- Self
- Function
- Repeat

Repeat actionblock's internal logic is still buggy, that was not part of this fix!

Additionally fixed fader suggestions, previously suggestions were not available for Fader element in code-block


Greg-Orca:
### passed

PR #1050, opened 2025-03-20
Fixed reactivity and syncing of mode blocks

### LINKED ISSUE #999: bug:  Mode block Button Mode field not working
**Describe the bug**  
The Mode field in the Mode Block does not send the correct value to the Grid. It is not possible to switch between modes.  

**To Reproduce**  

1. Open a Button Mode Block.  
2. Switch from Momentary to Toggle Mode.  
3. Press the button.  

**The button remains in Toggle Mode.**  

Additionally, if you switch back to Toggle Mode, it becomes Momentary Mode.  

**Expected behavior**  
The Button Mode field should work correctly when the user switches between modes.  

**Environment:**  

- OS: macOS  
- Editor version: 1.5.1 and the current nightly build

elsoazemelet:
Closes #999

Expected behaviour on all mode (Button, Potmeter, etc.) type blocks:

- Changing values (Input field, checkbox) trigger module activity (led lights pop, code is sent to grid)
- Sent changes are working as intended (see original ticket)


Greg-Orca:
### Test failed
Button Block Min Max is not working:
- Can't adjust Max parameter
-  when Min/Max enabled Max value jumps to 2

elsoazemelet:
> * when Min/Max enabled Max value jumps to 2

Fixed issues, could not repro issue above.

Greg-Orca:
### Test passed

PR #1052, opened 2025-03-20
Reworked and fixed Lookup block

### LINKED ISSUE #1010: bug:  Comma in Lookup 'Output' Variable Field Creates a New Input Pair Field
**Describe the bug**
When entering a comma (,) in the Lookup 'Output' variable field, the Editor unexpectedly generates a new input pair field.

**To Reproduce**

1. Open the Lookup configuration panel.
2. Navigate to the 'Output' variable field.
3. Enter a comma (,) in the field.
4. Observe that a new input pair field is generated.

**Expected behavior**
Commas should be allowed in the 'Output' variable field, but if the input is incorrect (e.g., improper use of the comma), an "incorrect input field" message should be shown, without generating new input pair fields.

- OS: mac os
- Editor version: 1.5.1

**Additional context**

https://github.com/user-attachments/assets/fbdde98d-adaf-4a2d-aa37-2bf09d4adb91

elsoazemelet:
Closes #1010 

Reworked lookup block to accomodate input fields and buttons from UI-kit.
All working is expected to be the same as before the change.

New expected working is similiar to variable type blocks:

- All entered values must be syntax correct
- Input fields with commas (and other invalid values) are considered syntax errors
- Lookup block is reverted to last syntax correct state after the block is out of scope

### DIFF
--- a/playwright-tests/data/actionBlockLocators.js
+++ b/playwright-tests/data/actionBlockLocators.js
@@ -17,7 +17,7 @@ export const blocks = (page) => ({
         input: page.getByPlaceholder("input").first(),
         output: page.getByPlaceholder("output").first(),
         destination: page.getByPlaceholder("Variable name to load the"),
-        addNewPair: page.getByText("Add new pair..."),
+        addNewPair: page.getByRole("button", { name: "Add New Pair" }),
       },
     },
     Global: {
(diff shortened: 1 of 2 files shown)


Greg-Orca:
### Test Passed
Lookup block working, and field checks works correctly

PR #1064, opened 2025-03-25
Fixed displayed character for international pound symbol

### LINKED ISSUE #886: Keyboard Macro Action wrong key label
Macro action has "?????? Keyboard Int' #" which is wrong, should be #/~/ű or something like that.

elsoazemelet:
Closes #886 

Displayed character for  0x32 HID (tested by changing keyboard layout in windows):
DE: # 
US: \\
HU: ű


Greg-Orca:
passed

PR #1068, opened 2025-03-28
🐛BUG: Added option for MonacoEditor for instance based element scope restriction

### LINKED ISSUE #1045: bug: monaco_elementtype is a single global store causeing invalid suggestions when switching element
**Describe the bug**
Tricky navigation can cause monaco to show incorrect suggestions in monaco modal!

**To Reproduce**

1. On a PBF4 module goto a potmeter element's timer event and open the code editor
2. type self:pot... and see that suggestions are correct
3. Navigate to button element using the dropdown menu in the configuration panel
4. Check the suggestions in the local variables action block on the button element's init event
5. Now if you click back to the code editor modal from the potmeter element suggestions will be incorrect (button suggestions are shown!)

elsoazemelet:
Closes #1045 

Bug description and repro steps can be found in original ticket.

Fix contains:

- Monaco editor suggestion creation is restricted on creation.
- Suggestions are now binded to element

To test: 

Try to change the element type while a monaco editor or LineEditor instance is present, and see if the monaco suggestions follow (drag and drop to different element, etc.).


Greg-Orca:
### Passed

PR #1102, opened 2025-04-11
🐛BUG: Fixed disabled store on ElseIf and fixed syntax error highlight

### LINKED ISSUE #1094: bug: Else if block blocking Store
**Describe the bug**
When using an Else if block, store is disabled

**To Reproduce**

1. Add an If block
2. Add inside an Else If block
3. Store disabled

- OS: macos
- Editor version: 1.5.3

**Additional context**

Image

elsoazemelet:
Closes #1094 

See original bug description in the linked ticket!

Additional fixes:

- Fixed display of blocks with syntax error (red border)


Greg-Orca:
### test passed

PR #1106, opened 2025-04-15
✨FEATURE: Elements are now navigable with keyboards when in focus

### LINKED ISSUE #1028: Module layout element focus naviagtion should be accessible via arrow keys, even between modules
## Element navigation using arrows!
Left, Right, Up, Down arrows in the middle panel should allow user to navigate between elements directionally!

Take into account possible module rotations! 

## Entering and exitting element context
Allow user to Enter module context (focus the config panel)
Or Escape config contect (focus back to selected element)

Image

elsoazemelet:
Closes: #1028 

See description in original ticket!

Added features:

- Module controll elements are now navigable with arrow keys when in focus.
- Elements can be selected between modules
- When current element has neighbouring element(s) with different dimension, the closest applicable one will be selected
- All the elements in any configuration should be traversable with keyboard
- Action lists can be entered via the focused grid element, when enter is pressed. This also works in Multi-Event mode, by focusing the first action list. Multi-Event mode focusing is not part of this PR.
- Action lists can be exited via Esc key.


Greg-Orca:
### test failed
- Can't navigate to system element by arrows
- If Preferences are open, and press Enter when an element on focus, then Editor Throws error
- Can't use esc when the cursor in a textbox

elsoazemelet:
> ### test failed
> * Can't navigate to system element by arrows
> * If Preferences are open, and press Enter when an element on focus, then Editor Throws error
> * Can't use esc when the cursor in a textbox

Fixed issues. The issue "Can't use Esc when the cursor is in a textbox" will not be fixed, as the focusing behavior of control elements in action blocks has not been implemented yet. This I currently consider a valid use case.

Greg-Orca:
If one of the modules is rotated (e.g., by 180 degrees), the navigation becomes incorrect — for example, pressing the up arrow will navigate down.
ezgif-560c76621c6620

Greg-Orca:
@SukuWc Could you please confirm how we should proceed regarding the issue: “Can't use Esc when the cursor is in a textbox”? Should we implement a fix or leave it as expected behavior?

SukuWc:
> @SukuWc Could you please confirm how we should proceed regarding the issue: “Can't use Esc when the cursor is in a textbox”? Should we implement a fix or leave it as expected behavior?

Escape behavior is acceptable as is for now, let's focus on improving arrow navigation!

Greg-Orca:
- Can’t navigate to the PBF4 buttons.
- When navigating to EF44 or PBF4, if the potmeter/encoder is the first element, it jumps to the fader.
- If EF44 or PBF4 is rotated, navigating from the fader always jumps to the potmeter/encoder.
video:
https://github.com/user-attachments/assets/353d1be2-fbd9-43b4-8857-8ee17aef60e3

Greg-Orca:
### Passed

PR #1107, opened 2025-04-15
🐛BUG: Fixed reverting to synced state and grid configuration sync

### LINKED ISSUE #1018: ux issue: deleting any actionblock value filed using backspace until invalid causes unintentional state when switching events
**Describe the bug**
When I delete an input field and switch between elements, the variable in the background remains the original one, but only the first character is displayed.

**To Reproduce**

1. Go to an input field
2. Click on and delete it by **Backspace**
3. Click on an element and switch back to the edited one.
4. The variable in the background remains the original one, but only the first character is displayed

**Expected behavior**
Last saved point should displayed properly, not just the last letter

- OS: macOS
- Editor version: 5.1.1 , 5.1.2

Image

elsoazemelet:
Closes #1018 

See bug description in original ticket!

The grid configuration syncing behaviour was fixed in the following way: 

- On change (clicking outside MeltCombo, etc.) the configuration is sent to module, if it is a valid config (no syntax error, no validation error). IMPORTANT, this may be broken in stable? This results in that the configuration on module is not present, only in editor. This is a worst case scenario, and may require a hotfix release.
- When an invalid action block is destroyed (due to being out of scope), it is reverted to last valid state, and sent to grid. This mechanic was broken, which resulted in the linked bug (#1018), and is fixed in this PR.

### DIFF
--- a/src/renderer/main/panels/configuration/components/DynamicWrapper.svelte
+++ b/src/renderer/main/panels/configuration/components/DynamicWrapper.svelte
@@ -68,7 +68,7 @@
     updateAction(
       action,
       new ActionData(action.short, action.synced, action.name),
-      false,
+      true,
     );
   }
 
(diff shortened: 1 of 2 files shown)


Greg-Orca:
### Test passed

PR #1125, opened 2025-04-30
✨FEATURE: Introduced Tour function

### LINKED ISSUE #945: Interactive getting started tour for profiles based on actionblock meta tags
Description: 
- User adds markdown description to a profile to describe the getting started steps
- When uploading the profile to the module, editor parses it to find corresponding custom action block names to jump to during the tour
- If action block name starts with "1." then that corresponds to the first getting started step.
- If multiple actionblocks have names starting with "2." then the tour jumps to each of them when pressing the next button in the tour popup. In this example color will be set by the user on all faders!
- The actionblock that the tour jumps to must open automatically
- Tour can interact with user_input directly to implement the navigation
- Buttons on the popup must be Next (Primary), Previous and Close
- No special formatting is required in the profile cloud!

Example markdown steps:

```
## Description
This profile does X Y Z
## Getting started
1. Edit MIDI channel on the Setup Event of the System Element
2. Pick a preferred color on the Fader element's Setup Event
3. Store configuration

```

Inspiration:

Image

elsoazemelet:
Closes #945 

To create a tour:

- Enable block names in editor settings
- Add blocks to your configuration with custom name, starting with a number, followed by a decimal point. (e.g.: 1. Getting Started)
- Save this configuration as preset/profile/snippet in Profile Cloud
- Modify configuration description in Profile Cloud the following way:
 ```
## Getting started

1. Edit the MIDI channel on the Setup Event of the System Element.
2. Pick a preferred color on the Fader element's Setup Event.
3. Store the configuration.

<!-- tour step=1 text[[Edit the MIDI channel on the Setup Event of the System Element.]] -->
<!-- tour step=2 text[[Pick a preferred color on the Fader element's Setup Event.]] -->
<!-- tour step=3 text[[Store the configuration.]] -->
 ```
 - Anything between a start and end tag will be considered as a tour step content. The number followed by step: will be used to match the step in the description (as in step:1:start) to the step defined in the blocks name.


elsoazemelet:
- [x] After loading a simple profile on a tour profile, the step blinks for a second
- [ ] ~~Fix Profile/Preset load overlay with load in progress indication~~
- [x] Add tour available button, it should start the tour and close the overlay (on overlay)
- [x] Add auto-open action block for tour step
- [x] Add lightweight invisible tour coding

elsoazemelet:
Calculation heavy operations (profile load) only freezes UI on virtual modules, since currently no rendering frame is available during process. This makes the optimization of calculate heavy runtime operations out of scope. Created separate ticket:

https://github.com/intechstudio/grid-editor/issues/1134

narayb:
When profile is dragged to module, tour icon does not appear when overlay was off. Overlay should appear when dragging to module.

Same types of blocks containing different types on information on the same event all get opened at the same time when the tour prompts it.
tour_not_exclusive_open

The text for guiding the tour should be in basic markdown, not in comment.

The ability to enter newlines would be nice. Sometimes Markdown uses doublespace to do this.

Tour step should follow action block position when closed

SukuWc:
- [x] Tour shoud be visible after restart if profile is present
- [x] Fetch all configuration from all connected modules before rendering load overlay
- [x] Connecting a second module while overlay is open causes runtime error
- [x] When two similar modules are connected loading and loaded states are incorrectly linked (common state)
- [x] Allow dragging and opening elements while tour is active, rerender tour popover to follow position
- [x] Last step of the tour, add Green Quit Tour button
- [x] Separate Loaded state display and Load button to allow Loading regardless of state
- [x] Start tour button should load the latest tour, this allows me to check the tour description and test the new tour without loading the profile to the module again.
- [x] FI from stable

kkerti:
I like the tour so far!

What I would love to see from this tech to support additional **key** anchor points. Lately as I've been working with configs, making them tidier, easier to understand, have less clutter and so on, for explanations I rely heavily on **element names**. Also upon suggesting for a customer to rewrite or delete a value somewhere, a **store** might be suggested. For example keybinding changes are not active until the configuration is stored. It would be neat to have anchor points for the Profile Cloud as well: Create a Profile | Upload a Profile | View the description | etc.

image

Greg-Orca:
### passed

PR #1135, opened 2025-05-14
Add quit dialog when user tries to close Editor with packages running in the background.

### LINKED ISSUE #990: When quitting Editor with Activated Packages, ask user if they want to minimize Editor to the taskbar instead.

danim1130:
If there is no running packages, nothing changes.
If there are packages running and the user tries to close the Editor with the "quit editor" option selected in Settings, a Quit dialog is shown, where a "minimize to tray" option is presented to instead minimize Editor.


Greg-Orca:
web build not working

Greg-Orca:
passed

Greg-Orca:
Modal: The Escape key correctly triggers cancel, but the Enter key also triggers cancel—it should trigger Quit instead.

"Close to Tray" isn’t ideal, as it's not a widely recognized or intuitive phrase. Consider using "Minimize to background" or simply "Minimize" to make more consistent with standard UI conventions.



In the future, if the feature is implemented  https://github.com/intechstudio/grid-editor/pull/1146, will users always see this modal message whenever the Debug Monitor or MIDI Monitor is activated?

danim1130:
Currently any enabled package will activate the dialog except for the built-in modules (debug/midi monitor). The code is ready for more sophisticated approach, we'll see how this behaves in real-world usage.

Greg-Orca:
### passed

PR #1138, opened 2025-05-15
✨FEATURE: Added Confirmation modal and reworked modal logic

### LINKED ISSUE #1118: Feature request: Codeblock modal escape and popup
**Escape Key / Cancel Button Behavior in Codeblock:**
When using the Escape key or Cancel button in the codeblock with active changes, the modal should not close immediately. A confirmation message should appear to prevent accidental closing of the modal.

**Escape Key Behavior with Monaco Code Suggestions:**
If a Monaco code suggestion is active, pressing the Escape key should first close the suggestion, not the modal.

elsoazemelet:
Closes #1118


Greg-Orca:
**failed:**
If a syntax error is present, the modal will not appear when pressing Esc or the Cancel button.
Modal is not responsive to Enter key

elsoazemelet:
> **failed:**
> If a syntax error is present, the modal will not appear when pressing Esc or the Cancel button.
> Modal is not responsive to Enter key

Fixed first issue!

About second issue of Enter not being responsive. Commented out the test for it, due to my concerns:

- Pressing Enter is often reflexive, and binding it to a destructive action like discarding unsaved changes could lead to accidental data loss.
- Editors usually default to safer behavior, requiring an explicit click.
- Discarding changes should be a conscious decision, not acting out of habit or speed.

To make this an expected behaviour, I need a second consideration and verification since this defeats the purpose of the popup modal.

SukuWc:
- [x] Please make the Enter behavior as described by the test
- [x] Re-add the test
- [x] When editing the code, whenever the code matches the previous committed value, then commit should be disabled and modal should not appear when hitting Esc.
- [x] FI from stabble branch

Greg-Orca:
After the modal closes, the cursor disappears, requiring the user to click back into the code editor to continue typing

SukuWc:
> After the modal closes, the cursor disappears, requiring the user to click back into the code editor to continue typing

@Greg-Orca we will not be able to fix this. This would require us to manually manage the cursor state for all of different HTML inputs. (overriding the default behavior)

Are there any others issues or can we merge?

Greg-Orca:
Passed

PR #1146, opened 2025-05-27
Rework navigation bar menu

danim1130:
The goal of this PR is to provide more emphasis for packages and move them to first-class components in Editor. This is achieved by reworking how packages interact with the UI. Instead of having a separate page to handle all package related things, each package gets its own place in the menu bar, which will display its preference component in the left side container.

To make the difference between packages and editor components as small as possible, the following steps have been made:
- Move Preferences into the left side panel container. With this change, every menu item will affect the left side panel.
- Remove the configuration panel menu item, since that is the only thing on the right side. Provide a floating button to show and hide the right side panel.
- Move the Debug Monitor, Midi Monitor, Profile Cloud and Websocket Monitor into the Packages list as toggleable elements. This will change which menu item is available.
- For each package enabled, show a menu item if possible. 

Note that currently, the web build **does not support** the package system, and so the menu system is hardwired to only contain the existing components.


Greg-Orca:
Please disable the scrolling vertically
Screenshot 2025-05-29 at 15 08 10

Greg-Orca:
https://github.com/user-attachments/assets/0e9869af-63f0-42e1-895c-e17b0dd6ea7c

Hover text would be a nice feature, especially if users are using more than one package. Text is only visible when the icon is clicked.

danim1130:
In the future, every package should have it's own icon, the general icon is just a fallback. For this reason, I have modified the code so that the toolbar is shown in hover only if the package does not have a unique icon.

Greg-Orca:
Could you please add this hover function to all the icons so that the function is complete and consistent? (Package, Preferences)

Screenshot 2025-06-02 at 11 55 41

danim1130:
Done

Greg-Orca:
When I hover over the preferences icon on the right side, the cursor rapidly switches between the hover and non-hover states. This causes the text to blink and makes the icon non-interactable.

ezgif-829f8f94370690

danim1130:
Fixed by reworking how the tooltip is shown and hidden

Greg-Orca:
**passed**

Greg-Orca:
**passed**

PR #1158, opened 2025-06-16
🐛BUG: Fixed focus border when selection is present on element

### LINKED ISSUE #1131: Track Element does not move the blue focus rectagle properly

elsoazemelet:
Closes #1131

Focus border now moves with element selection when selected element is changed due to interaction on physical module.
image

**IMPORTANT!** 

Fixed tracking, currently on stable event/element tracking is unresponsive.


Greg-Orca:
**passed**

PR #1172, opened 2025-06-25
✨FEATURE: Replaced send inmediate input to Monaco based editor

### LINKED ISSUE #1132: Immediate send text imput in debug sidebar should be monaco editor based!

elsoazemelet:
Closes #1132 

Replaced send inmediates input field to a monaco based mini-editor in Debug Monitor. 

Added simple LUA language for send immediate auto suggestions. This panel now does not suggest Grid specific functions.

**NOTE:** Monaco editor should still suggest the same Grid + LUA specific functions correctly!

### DIFF
--- a/src/renderer/lib/monaco.ts
+++ b/src/renderer/lib/monaco.ts
@@ -38,7 +38,7 @@ const language_config = {
   ],
 };
 
-export const language = {
+const [lua, intech_lua] = Array(2).fill({
   defaultToken: "",
   tokenPostfix: ".lua",
   keywords: [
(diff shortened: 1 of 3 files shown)


SukuWc:
- remove all grid specific suggestions. (only keep print, if, math.floor etc)

Greg-Orca:
bug: I tried both the print function and the potmeter centering, but it looks like they’re not working.

note: If you navigate away from the Debug view, everything you’ve written in the field is lost.

elsoazemelet:
> bug: I tried both the print function and the potmeter centering, but it looks like they’re not working.
> 
> note: If you navigate away from the Debug view, everything you’ve written in the field is lost.

Tested with the following commands: led value, print, and midi. It seem to works. Tested on local branch and on hosted web view.
The value is lost since the debug monitor is out of scope. Consulted with @SukuWc , it was the working of the send immediate panel before, and this working is to be kept (not an issue).

Greg-Orca:
Passed

PR #1182, opened 2025-07-02
Modernize package list UI

danim1130:
Modernize Package Manager list view to make it easier on the eyes and improve usability.
This PR contains a number of changes to the Package Manager:
- Changed the handling and ordering of packages.
- Handle different transients states, such as package updating
- Load package data from Github
- Cache Github package info


kkerti:
Please do a short brief in your PR's dear developer! 🗡️

### Things discovered, noted in web version
- Preferences is not the most important Editor setting to look for. Either put it as last menu option in the left bar, or put it left bottom. See VSCode for inspiration: 
image

- Clicking over the packages (incl. debug, midi monitor), the different panels have different in and out transitions. When clicking them in different order, they have a different load transition. Consolidate them to either 
  - a.) no transition 
  - b.) transition **in** at X speed, but transition out immediately.
- Add a hover effect for the arrow button which hides the right configuration panel. (I love that arrow & feature btw!)

### Things noted in downloaded version
- I really love the hover tooltips! 🪄
- What is this big blob of text, when clicking on the Active Window package for example? _I assume it's a debug log, but please give then a name for this part_
image

- It is hard to know **when** the Active Window works, **or if it actually does something**. The only way to check if this package does something (without any code, docs lookup, etc.) is looking at a very small text, in the Editor. Probably this text is not visible, as a different window floats over Editor too. **We should give the user confidence, that after few of their clicks, something happens.** Ideas for this:
  - Give an indication that the package is "connected", just works by adding a green dot next to the package icon.
  - Whenever an active window is changed, the detected window title comes up with the help of the overlay feature (similar to PS). This should be a toggle option for the active window package, by default may be turned on. 
image

- (design needed 🎨) The package manager buttons should have different priority. Right now all buttons look like the most important thing to click. There are contra-dictionary actions next to each other like _jump of this cliff_ or _eat ice cream_, both marked as priority. I see that this comes from VSCode's extension manager, **BUT** there when you don't have an extension installed, your only action can be "install". If the extension is already installed, then from left to right, more serious actions follow each other, like the lighthearted restart, the annoyed disable and then the very angry uninstall. Here some design help should be asked for from the team.
- (design needed 🎨) It is very hard to understand and separate, which button (enable/disable,etc...) relates to which package. Need more padding, or border-b, something
image

- (design needed 🎨) We should revisit what should be written to the top which is now Packages, followed by the _Installed packages_  and the the _Available packages_. As this PR came to realisation, (which is very cool), we have to revisit these details at least in copywriting and UI arrangement to make the tech features a joy to use.

- Package already installed and then enabled won't fetch the fancy new details. Have to remove it first and add it back. Maybe we should show that there is higher version available of the given package?
image

- THE PACKAGE INSTALL FLOW WITH THE LOADING BAR IS EPIC! ✨

danim1130:
- Copy vs code extension view with buttons, only show update button for installed packages when it is available, context menu otherwise

Greg-Orca:
Checkbox megtévesztő

It looks clickable and confusing. If we really want checkboxes, then use this standard Blue one

Screenshot 2025-07-21 at 14 37 14

Greg-Orca:
looks nice
Passed

PR #1184, opened 2025-07-03
🐛BUG: Fixed focusing behaviour for Element selection both for single and multi-connection

### LINKED ISSUE #1164: bug: No Element focus border on Additional ports
**Describe the bug**
If the user is using multiple ports, the element focus and navigation feature only works on the first port, not on the others.

**To Reproduce**

1. Connect two modules on separate ports
2. Turn on Element Track
3. Go to the **second port**
4. Interact with an element
5. **No Blue Border focus!**
6. Click on an element
7. Border Focus appears
8. Try to navigate with arrows
9. **Can't navigate**

- OS: macOS
- Editor version: 1.5.8

**Additional context**
https://github.com/user-attachments/assets/c59fc3b9-9ff3-43ff-871a-9daa92ba5dc1

elsoazemelet:
Closes #1164 
See original ticket for detailed bug description.


SukuWc:
Please allow opening and closing the minimap pane even when only one setup is connected. Use the same arrow button that is used by the configuration panel! With one setup default state should be closed.

Greg-Orca:
@SukuWc 
Issues to discuss:

 - When switching between ports, the selected element appears random and does not remain where it was previously.

 - The Tab key does not work in the Minimap pane — it only tabs to the first port and does not move focus to other ports.

elsoazemelet:
> @SukuWc Issues to discuss:
> 
> * When switching between ports, the selected element appears random and does not remain where it was previously.
> * The Tab key does not work in the Minimap pane — it only tabs to the first port and does not move focus to other ports.

Fixed tab focusing. The selected element is always Element 0 when switching ports.

Greg-Orca:
### Passed

PR #1188, opened 2025-07-07
✨FEATURE: Added Reset Encoder action block

### LINKED ISSUE #1175: Feature: Reset Encoder Rotation actionblock
This simple action block allows user to set the rotation value of an encoder element. (Usually called from the button event of the encoder)

Reference design:
Image

Fields:
1. Element field behaves exactly like it does in simple color action block. This selects the element or self as the scope for where to run all of the functions

2. Value field expects a numeric value or expression, the result will be set as the rotation value for the element. Suggestions must be as follows:
- text: Zero, value 0 (number) 
- text: Minimum, value: "0%" (string)
- text: Center, value: "50%" (string)
- text: Maximum, value: "100%" (string)

Note that percentage based values are not yet implemented in firmware

3. Trigger event automatically is checked by default. If unchecked self:get(2) must NOT be added to the action string

Default lua:
```self:eva(0) self:get(2)``` 

self:get(2) means event trigger on the current element, parameter is the event type (2 = GRID_PARAMETER_EVENT_ENCODER)

elsoazemelet:
Closes #1175 

See original ticket for feature description!


Greg-Orca:
Values with percentages are not working — they always reset to 0.
Screenshot 2025-07-10 at 10 32 09

Please hide the Automatic Mode in both Minimalist and Advanced modes.
If the user wants to create more complex configurations, they can use the Code Blocks instead.

elsoazemelet:
> Please hide the Automatic Mode in both Minimalist and Advanced modes.
> If the user wants to create more complex configurations, they can use the Code Blocks instead.

@SukuWc can you affirm this? The auto-trigger was specifically part of the design.

Greg-Orca:
- [ ] ~~Values with percentages are not working — they always reset to 0.~~

- [x] Can't pick reset encoder block on the Button event, only on the encoder event!

elsoazemelet:
> * [ ]  ~Values with percentages are not working — they always reset to 0.~
> * [x]  Can't pick reset encoder block on the Button event, only on the encoder event!

Reset Encoder is now not event type specific. After thinking it through, any part of a module should be able to reset any encode in the setup.

After discussion with @SukuWc the encoder reset in percentages are not yet implemented.

Greg-Orca:
If I assign this to the encoder to button and press it, it triggers the encoder and sends out the reset value.
However, when I assign it to another element, it doesn't send out the reset value immediately.

@SukuWc  How should this work?
Maybe it should always trigger the encoder regardless of which element is used.

SukuWc:
Blocked by required firmware change!

Issue #1137, opened 2025-05-15
Package Manager UI rework, enhancements, finalization


danim1130:
## Package Manager improvements

- [x] Rework menu system to accomodate packages
- - [x] Move settings to the left panel, make the right permanent configuration panel
- - [x] Let packages define menu items, move preferences to this page, add package debug log here
- - [x] Consider adding there Debug and MIDI Monitor as "default packages"
- [ ] Rework package list page
- - [x] Status info, ready, up to date, progressbar for download and zip
- - [x] Mark intech packages as special/recommended package
- - [x] Add icon to package left position to name
- - [x] Adding / reinstalling package should restart the package manager automatically (installed package should work right away)
- - [x] Check if node process remains active for package side-loading while shutting down editor
- - [ ] Check downloaded or updated package folder name on release (specially with mac!)
- - [x] Make packages have fix positions in the list, dont jump them around based on current status
- [x] Integrate actions into packages closely
~~- - [ ] Allow enabling package from action picker, show what actions are provided even by disabled packages~~ Idea declined, redundant.
- - [x] Support ~~short code~~ category search for actions ~~(ps ... or lr ...)~~ (Photoshop or Lightroom)
- - [x] Unknown action should enable needed package, not convert to code block (at least for intech packages)
- [ ] Getting started with packages
- - [ ] First time user welcome modal, help installing packages for primary workflows (audio, photo (ps/lr...), etc.)
- - [ ] (Tofi: use guided tour or else to support onboarding for package usage, newcomers)
- [ ] Integration with Profile Cloud ~~(ask to automatically enable package)~~
- [x] Package manager extras
- - [x] ~~Enable right click interactions on package icon navbar~~
- - [x] Show download progress in menu 

## Package functions
- [ ] Overlay
- - [x] Decide on overlays supported by packages, figure out common solution
- [ ] Spotify
- - [x] Decide Spotify release mode (with current developer mode plugin and rate limit / everyone creates spotify app (like discord))
- - [ ] Finalize Spotify visuals on VSN
- - - [ ] How to render album titles with unicode characters?
- - - [ ] Album image handling
- - [ ] Cache next/previous song to make switch instant on module (keep it on the package)
- - [ ] ~~Add public callbacks for users?~~ Probably later
- - [ ] Authorization callback success page? (Currently closes tab)
- - [ ] Code refactor (index.js is too long)
- [ ] Subscribe to changes
[...]

PR #1206, opened 2025-07-21
Look for category name in action picker search as well

danim1130:
Based on one point of https://github.com/intechstudio/grid-editor/issues/1137

The goal of this PR is to extend the search functionality of the action picker to also include the category of the actions, such as showing all Photoshop actions when any part of "photoshop" is in the search terms.


Greg-Orca:
### Passed

PR #1229, opened 2025-07-30
Show enable package instead of converting to code

### LINKED ISSUE #1225: Raw block should be able to recommend the required package to be installed

danim1130:
The goal of this PR is to make it possible for the user to enable an installed package from a missing action block.

To test, have a package that adds actions to the Editor (such as Spotify, Photoshop, Lightroom). Enable the package, add the action to an element, then disable the action. The missing action block should be shown, with an enable button that turns the package back on.

### DIFF
--- a/src/renderer/config-blocks/RawCode.svelte
+++ b/src/renderer/config-blocks/RawCode.svelte
@@ -79,7 +79,7 @@
       event: "Package Manager",
       payload: {
         click: "Status Change - Raw Code",
-        id: packageId,
+        id: availablePackage.id,
         status: true,
       },
       mandatory: false,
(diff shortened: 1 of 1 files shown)


Greg-Orca:
If I uninstall the package, the Codeblock still shows 'Enable Package' and does nothing.

In this case, it should display 'Install Package' and trigger the installation when clicked. Once the installation starts, it should either show a progress indicator or and grey out the button—some kind of visual feedback to show that something is happening

Greg-Orca:
### Discord, PS, Spotify enable module with block throws error
Screenshot 2025-07-31 at 16 08 16

### Lightroom package shows "To Code" button
Screenshot 2025-07-31 at 16 12 21

danim1130:
Exception is fixed

Lightroom has to be updated, it should work now on the latest lightroom package version

Greg-Orca:
Passed

PR #1236, opened 2025-08-04
Fix LedColor color picker freeze

### LINKED ISSUE #1235: bug: Color block color change freezes interactions on Release
**Describe the bug**
Adding the original Color action block to a module and changing the color value freezes all interactions with the UI (app remains responsive, package runs).

**To Reproduce**

1. Add Color action block
2. Change color value
3. See error in console log

**Expected behavior**
Color is set correctly

- OS: Windows for sure, probably all
- Editor version: Latest Nightly

**Additional context**

Image

Caused by lib/_utils.js:

```
  export function parseRGB(r: any, g: any, b: any): RGB | undefined {
    if (![r, g, b].map((e) => parseInt(e)).every((e) => Number.isFinite(e))) {
      return undefined;
    }

    return new RGB(parseInt(r), parseInt(g), parseInt(b));
  }
```

RGB class has been removed in LightMode PR #1210, but this function remained.

danim1130:
Fixes https://github.com/intechstudio/grid-editor/issues/1235

### DIFF
--- a/src/renderer/config-blocks/LedColor.svelte
+++ b/src/renderer/config-blocks/LedColor.svelte
@@ -104,6 +104,14 @@
     handleConfigChange($config);
   }
 
+  function parseRGB(r: any, g: any, b: any): Color.RGB | undefined {
+    if (![r, g, b].map((e) => parseInt(e)).every((e) => Number.isFinite(e))) {
+      return undefined;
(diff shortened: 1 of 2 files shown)


Greg-Orca:
passed

PR #1240, opened 2025-08-12
Fix profile load race condition for profile cloud

danim1130:
Clicking between presets and profiles should not freeze UI.

pc_freeze


kkerti:
Profile, snippet and  element preset load works OK.

PR #1243, opened 2025-08-14
Reworked element naming

### LINKED ISSUE #1232: Feature: Add pinned element name field to configuration panel
## Reference design
Image

## Description
If the selected element's setup event's first actionblock is an element name action block then...
- The two fields should be bidirectionally synchronized
Otherwise
- The pinned element name field should be empty.
- If the user writes into the pinned field then the element name action block should be created on the seleted element's setup event

## Element name actionblock improvements
- Add 'Allow expression' checkbox
- When this is checked then `"` character should not be added to the generated short code
- Element name actionblock should be and 'alwaysopen' actionblock like the commetn block

elsoazemelet:
Closes #1232

See original ticket for feature definition.

Element name detection and changing is reworked.

- When an element name is changed via the new Element Name field:
    - If no Element Name block is present on setup event it is added. The value of the input field and the block should synchronously update each other
    - If Element Name block is present, it is updated
- When an element name is changed via the Element Name block:
    - If newly added manually, the name should update on the input field
    - If already present, changing it should update the input field
    - If deleted, or moved from first position, the input field resets

**NOTE:** Previous grid to editor element name runtime calculation is ignored (and is not used anymore)


Greg-Orca:
The element name is not cleared/changed after:
- delete the setup>element name action block
- clear the element name
- only add a "space" character

Greg-Orca:
Screenshot 2025-08-27 at 11 37 34
The element name is not cleared if I delete the element name block,

Greg-Orca:
Screenshot 2025-08-28 at 11 25 15
please add a test id for this input field

elsoazemelet:
> Screenshot 2025-08-28 at 11 25 15 please add a test id for this input field

Added ID, fixed locator. Some tests still are broken, can you please check on them? @Greg-Orca

Greg-Orca:
please use data-testid instead of data-test-id

https://playwright.dev/docs/locators#locate-by-test-id

PR #1304, opened 2025-09-11
Reworked firmware update process and design

### LINKED ISSUE #309: DESIGN: Visualize Grid when plugged in boot-loader mode
ASK @elsoazemelet  about design prototypes

TECHNICAL STORY:
Currently when a module is connected in boot-loader mode, a message is displayed in the notification bar. Make it more visual. Propose design first.

This would greatly increase the UX of firmware updating on these devices. Also this makes it clear that when a device is plugged into Editor (either normally or in bootloader mode) it will appear as that device on the screen - indicating that it is in fact, plugged in.

DESIGN:

Image

## Implementation plan
- [ ] No animation just color change on svg!

elsoazemelet:
Closes #309 

See original ticket for design and functional requirements!

The normal and nightly update process should work as previosly both for Grid and Knot.
The update states should be correct at all times, and should work after if disrupted at any stage.

# Example Use Cases:

```
1. Connect FW mismatched module
2. Proceed with normal/nightly update
```
```
1. Connect non-FW mismatched module in bootloader mode
2. Proceed with normal/nightly update
```
```
1. Connect non-FW/FW mismatched module in bootloader mode
2. Interrupt the update process
```

etc.


Greg-Orca:
Failed. After the FW update, the editor always shows an unsuccessful message, but the update is actually successful.

tested on: MacOS

elsoazemelet:
> Failed. After the FW update, the editor always shows an unsuccessful message, but the update is actually successful.
> 
> tested on: MacOS

Update/Fail state comes from the electrons built in updaíte procedure. This should not be in connection of this PR. Needs reassuring on MacOS. Note, on windows, update procedure works as normal. @SukuWc @kkerti

kkerti:
### Modal not shown on second fw update mode connection.

1. After entering fw update mode first time, update modal shows as it should.
2. Clicking on "Dismiss", if I connect the module **again** in fw update mode, the modal is not visible.
3. I have to connect the module once "normally", to again disconnect and connect in fw update mode to see the modal.

### Update unsuccessful after successful update (?)
Issue remains as of today's test (grid-editor-macos-1.6.1-nightly.1.20250915.1550-arm64), where the last info on the modal is unsuccesful update, but in reality the update completes fine.

elsoazemelet:
image

elsoazemelet:
Update state 6 (failed update) is always shown, since when USB connection is lost at the end of update process, the copy OS function fails, and the error state is set. This working of FW update is shared with previous versions, and can not be fixed currently. 

Solution: Reworked communication of the update process

elsoazemelet:
- [x] Unstable states:

1. Connect in bootloader mode
2. Dismiss
3. Reconnect in bootloader mode
4. Update dialog is not shown

- [x] Add dismiss to firmware mismatch notification bar
- [x] Make it stylish in light mode too!

Greg-Orca:
passed

PR #1308, opened 2025-09-11
Fixed in/out coming message rate display NaN issue

### LINKED ISSUE #1305: TX data rate numeric value is not displayed correctly in debug monitor
Image

- [x] Sometimes NaN is shown even after connecting a module
- [x] Sometimes good value is shown but turns to NaN after fetching some events

elsoazemelet:
Closes #1305


Greg-Orca:
can't reproduce the NaN

passed

PR #1309, opened 2025-09-12
Multiple configuration panel QoL fixes

### LINKED ISSUE #1286: bug:  Action list minor usability fixes
## Todo
- [x] Action list scroll bar should be displayed at the right-most edge of the panel without padding
- [x] Make sure scrollbar is correctly position in MultiEventView
- [x] Action blocks should not be draggable using right mouse button
- [x] Action blocks should not be draggable when Ctrl + LeftClick because that is used for Selection only!

Image

Image

elsoazemelet:
Closes #1286 

See original ticket for detailed description of required minor fixes in Configuration panel.


Greg-Orca:
passed

PR #1313, opened 2025-09-16
Fixed element name calculation

### LINKED ISSUE #1220: bug:  Element name is only updatede on overlay if another element is selected
## Repro steps
- Connect a real module
- Clear the configuration
- Open the Element Name overlay
- Enter a name into the new element name for the selected element (name is not displayed on the overlay)
- Select another element using the dropdown in the configuration panel (name is still not displayed on the overlay)
- Select the original element from the same dropdown menu (name is now displayed on the overlay correctly)

## Bonus item
- [x] When I clear the module configuration element name store should also be cleared
- [x] Same for Discard All
- [x] When I clear the element using Clear Element from the context menu then that element's name should be cleared
- [x] Same for element discard
- [x] Receiving EVENTVIEW class should also update the element name store

elsoazemelet:
Closes #1220 


Greg-Orca:
Bug: After clearing the whole module, the element name is still displaying


Screenshot 2025-09-25 at 11 40 22

elsoazemelet:
https://github.com/user-attachments/assets/3f68aaa0-7f82-4bf4-9ea2-e7e255e044c6

Reprod. on web preview and local branch, element name is cleared on module clear when element name is stored and when it is in memory only. Also tested with virtual modules.

Needs further repro steps!

elsoazemelet:
Repro steps: Rename 3 elements

Greg-Orca:
passed

PR #1314, opened 2025-09-16
Fixed MeltCombo propagation and light mode colors

### LINKED ISSUE #1288: UIKIT: MeltSelect dropdown stays open when clicking on MeltCombo
Image

## Bonus:
- [ ] MeltCombo should use the ssame light mode colors as MeltSelect

elsoazemelet:
Closes #1288 

**Note:** MeltCombos click event now propagates when the input field is clicked. Not sure why previously it did not propagate, possibly by mistake. Tested with keyboard navigations too, it does not make a difference now, but mind that, MeltCombo might be tested (if there is other ideas to test it) if it remained the same.


Greg-Orca:
fixed

SukuWc:
@elsoazemelet please check this again, there seems to be some issue

Greg-Orca:
passed

PR #1318, opened 2025-09-22
Introduced rename feature for all users, reworked action block component API

### LINKED ISSUE #1290: bug:  Missing Edit Name "Pencil" icon for Midi block
Image

Fix for the following action blocks:
- [x] MIDI
- [x] MIDI14
- [x] MIDI NRPN
- [x] MIDI SYSEX

Todo:
- [x] Check all other action blocks to make sure pencil is available
- [x] Remove editable block names option from preferences and make enabled for all users
- [x] Add 'Edit actionblock name' as an option to the context menu on actionblocks. this should open a simple  modal to do the name editing!

elsoazemelet:
Closes #1290

See original ticket for feature description!

Currently all blocks are renameable, except composite blocks (e.g.: If, for loop, encoder left/right, etc.). These in my opinion should not be renamed, since they are functional blocks (logical operators).

Additionally Action Block API  was changed. Data field "config" and all connecting functions are renamed to action, and all instances were typed to GridAction/ActionData correctly and accordingly. Since many packages are using the same interface, they are not refactored to use action instead of config export property (yet). 

However, this was handled in a backward compatible way, so it should not throw runtime error, or warning. This was commented accordingly, and DynamicWrapper was refactored to accomodate these changes. 

**NOTE:** Tested with Lightroom package


elsoazemelet:
- [x] Detect during runtime if config or action prop is available!

Greg-Orca:
Bug: If you change any parameter on a renamed action block, the action block’s name is deleted.

Greg-Orca:
passed

PR #1320, opened 2025-09-24
Add calls for packages to support configurator mode

danim1130:
The goal of this package is to provide the ability for packages to also use the minimalist mode flag in the action configurator. Package support is also needed.

To test, use the latest Lightroom package (1.0.3) and add the Develop Control action. When switching to the minimalist mode, only the parameter name should be editable. Relative-absolute mode can be set on the corresponding mode action.


Greg-Orca:
@SukuWc we should find a better solution. Relative mode is really important at the moment (since there is no bi-directional communication, Absolute mode is not usable).

Currently, the user needs to guess which min–max range to use for each type of parameter. I think we should handle this on the package side, so each function knows how precisely it should adjust each parameter

danim1130:
Updated Lightroom package based on discussion

PR #1322, opened 2025-09-25
SUKU tester enable checkbox added to preferences dev settings

### LINKED ISSUE #1303: BUG: Editor hijacks MIDI/USB device
Tested on WIN10.

Steps to repro:
1. Have Editor running and open the MIDI monitor Debug View (normal MIDI monitor might ven be enough)
2. Start a DAW software
3. See if Grid is recognized as a working MIDI USB device (in ableton is looks like this when it's not)

Image

4. profit

Editor should never hijack devices like this.

Currently the fix is to close both software and start the DAW first to let it have the device to itself and start Editor after.

SukuWc:
closes #1303


narayb:
Seems good!

PR #1325, opened 2025-10-01
SUKU fix constant polling of element name

SukuWc:
- Editor kept requesting the element name after each interaction, this caused missed heartbeat timing, now fixed.
- Editor incorrectly reset the element name to empty string, now this is fixed too
- Editor now properly listens to EVENTVIEW messages and decodes name parameter from them


Greg-Orca:
passed

PR #1346, opened 2025-10-17
SUKU skip element name requesting for system element

SukuWc:
@Greg-Orca originally reported the following issue:

image


Greg-Orca:
passed

PR #1378, opened 2025-11-30
Clear to custom defaults

SukuWc:
image


Greg-Orca:
The process works, but it’s very slow because it first saves the background the current config, then uploads the new one. We should skip the background save, it should upload the profile immediately, since this action is essentially a ‘Clear’
use this to test:
Screenshot 2026-01-08 at 14 13 07

Greg-Orca:
After Loading a profile from the profile cloud, the store button is not available!

SukuWc:
@Greg-Orca fixed

Greg-Orca:
passed

PR #1395, opened 2026-01-02
Migrate minification to stylua

SukuWc:
- New grid-protocol version uses wasm-fmt/lua_fmt package (based on stylua)
- Inline comment's are now handled correctly (newlines are preserved correctly when needed)


Greg-Orca:
if you make errors:
Screenshot 2026-01-08 at 14 21 51

Greg-Orca:
### Comment bug

#### Setup

Use two lines in the code:

```lua
self:midi_send(0, 176, 1, 1)
self:midi_send(1, 176, 2, 2)
```

#### Steps & Results

1. Check the MIDI Monitor → **both MIDI messages are sent** ✅
2. Comment out the **first line** and commit  
   - Validate → **only the second message is sent** ✅
3. Uncomment the first line and commit  
   - **Both messages are visible again** ✅
4. Comment out the **second line** and commit  
   - ❌ **Error is thrown** (see screenshot)  
   - ❌ **Both messages are still sent**, but it **should send only the first line**
   Screenshot 2026-01-08 at 14 30 49

5. Store the config  
   - ❌ **Same issue persists**
6. Comment out **both lines**, then commit and store  
   - ❌ **Both messages are still being sent**

#### Expected behavior

- Only uncommented `midi_send` lines should be sent
- Commented lines must not produce MIDI output

SukuWc:
proposed fix: must add newline at the end of minified code if the last line of minified code is a line comment

Greg-Orca:
The comment broke when the module was replugged.

SukuWc:
Error field should be at lease 1 line high even if empty

PR #1437, opened 2026-03-02
Add EVALUATE debug panel

SukuWc:
## Firmware
Must use Nightly Firmware

## Summary

- New **Evaluate** tab in the Debug Monitor panel: Monaco editor for Lua code, sends a Grid `EVALUATE` frame and displays the typed response (ACK/NACK, message ID, decoded Lua return values)
- New `evaluate-parser.ts`: decodes EVALUATE response elements — nil, boolean, number, string, and nested tables
- New `SendRaw` component extracted from `SendImmediate` into its own panel
- `sendRawDataToGrid()` added to `WriteBuffer` — queues pre-encoded raw bytes through the engine FIFO with full `responseRequired` / `filter` / `responseTimeout` support; replaces the old `sendRawToTransport()` bypass
- Evaluate is now the default selected panel; module selector lifted to `DebugMonitor` parent and shared across all send panels

## Test plan

- [x] Connect a Grid module and open the Debug Monitor
- [x] Select a specific module from the dropdown (or leave as Global)
- [x] In the **Evaluate** tab, enter `return 1, "hello", true` and click Send — verify ACK status, message ID, and decoded values appear
- [x] Enter `return nil` — verify `[null]` is returned
- [x] Enter a table: `return {a=1, b=2}` — verify table is decoded correctly

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.6.5",
       "hasInstallScript": true,
       "dependencies": {
-        "@intechstudio/grid-protocol": "1.20260131.1058",
+        "@intechstudio/grid-protocol": "^1.20260302.1321",
         "@intechstudio/grid-uikit": "^1.20260206.1447",
         "@intechstudio/profile-cloud-webcomponent": "1.20251107.1414",
         "adm-zip": "^0.5.10",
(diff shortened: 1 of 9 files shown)


Greg-Orca:
passed

PR #1459, opened 2026-03-18
SUKU File Explorer panel initial implementation

SukuWc:


SukuWc:
Firmware (grid-fw):                                                        
  - FETCH handler now sends NACKNOWLEDGE for all recall_configuration failure
   cases (inactive page, nvm busy, invalid element) instead of silently      
  dropping the request
                                                                             
  Editor (grid-editor):                                                      
  - NACKNOWLEDGE added to InstructionClass enum
  - validate_incoming now detects NACK by checking class_instr first, matches
   on BRC + class_name + LASTHEADER, and calls waiter.destroy() to terminate
  the request as an error (no more infinite retry)                           
  - FetchConfig filter gets LASTHEADER: null for proper NACK correlation

PR #1463, opened 2026-03-26
Speed up both web and electron app startup

kkerti:
- Fixed a regression caused by untargeted tailwind blob parsing. 
- Monaco now only imports the core editor, without the various unused language modules (python, rust, etc...)


kkerti:
I have decided to merge this PR into https://github.com/intechstudio/grid-editor/pull/1461, as it's a very small change, but still monaco editor is impacted by it. No need to test separately this way.

PR #1470, opened 2026-04-02
LuaLS as language server for Monaco editor

SukuWc:
## Summary
- Integrates LuaLS (Lua Language Server) with the Monaco editor via a WebSocket bridge
- Spawns LuaLS as a child process on the main thread, bridges stdio ↔ WebSocket for LSP JSON-RPC in the renderer
- Adds `---@meta` annotations for the Grid Lua API (`build-assets/lua-annotations/grid-api.lua`) enabling completions, hover, signature help, and semantic tokens

## Test plan
- [ ] Verify completions appear for Grid API globals in the Monaco editor
- [ ] Verify hover and signature help work on function calls
- [ ] Verify LuaLS binary downloads correctly via `postinstall` on all platforms (macOS arm64/x64, Windows, Linux)
- [ ] Verify app builds and packages correctly with LuaLS binaries in `extraResources`
- [ ] Verify CI build matrix includes the LuaLS download step


Greg-Orca:
In this PR, if I disconnect the modules immediately after connecting them, an error occurs.

Screenshot 2026-07-01 at 15 48 26

Greg-Orca:
code editor on element no language server
Screenshot 2026-07-01 at 12 33 47

kkerti:
> In this PR, if I disconnect the modules immediately after connecting them, an error occurs.
> 
> Screenshot 2026-07-01 at 15 48 26

This won't pose a problem on production release branch. @SukuWc reworked how console logs are shown in Error .svelte component, and some errors are surfaced in that top error area. These are for extra heads up during development and testing.

In this case, it's expected to see these messages, indicates that after disconnecting the module, the background process which loads in the config from the module still sent out some messages after the disconnect, but they weren't able to return as the module was already disconnected.

kkerti:
> code editor on element no language server Screenshot 2026-07-01 at 12 33 47

Now lua language server should be initialized.

kkerti:
- Overflow is now allowed for the monaco widgets
- LineEditor has the same context awareness as CodeEditor
- Hover and Suggestions are **not** disabled, instead the widget overflow helps in visualization
- Added extra context check for right hand variable declaration in VariableManager

Action blocks impacted:
- [ ] Self
- [ ] Local
- [ ] Global
- [ ] If
- [ ] Else If
- [ ] CodeBlock (inline)
- [ ] CodeBlock (modal)

Greg-Orca:
command+S save
Screenshot 2026-07-03 at 14 27 18
Screenshot 2026-07-03 at 14 28 38
Screenshot 2026-07-03 at 14 30 08

kkerti:
> command+S save Screenshot 2026-07-03 at 14 27 18 Screenshot 2026-07-03 at 14 28 38 Screenshot 2026-07-03 at 14 30 08

1. Regression since vscode-api handled save differently
2. added `lualsExpressionOnly` flag
3. extended openEditorContext with self. typings

kkerti:
Fixed animation lifecycle, so hover info works properly and doesn't do an invalid size calculation between transition frames.

Greg-Orca:
- cloud be nice if we disable Global variable check ---@diagnostic disable: undefined-global

- resize bug:
Screenshot 2026-07-06 at 12 39 46

- no element suggestion
Screenshot 2026-07-06 at 12 40 13

- after `self.` gives suggestion:
Screenshot 2026-07-06 at 12 40 36

kkerti:
### fixed:

- `global` diagnostics are disabled now, until we figure out how code contexts work well with Grid
- undefined `inject-field` diagnostic is disabled, same story as with `global`
- regression fix: element suggestions work now
- regression fix: icons are shown next to suggestions
- regression fix: monaco resize works now

### won't fix:
The _valid_ suggestions are rendered first, and then the invalid ones later.

Not valid suggestions by suggestion provider and provideDefinition for self. and self: (dot and colon) will be reworked with the completion provider in separate PR, possibly with the context extension work.

kkerti:
Regression to fix: file manager panel should load luals or fallback server on init.

PR #1495, opened 2026-05-19
SUKU non-configurable special action blocks refactored into factory pattern

SukuWc:
Test the following in the UI:
1. Open an encoder element → verify encoder blocks show, endless blocks don't
2. Open an endless element → verify endless blocks show, encoder blocks don't
3. Add an encoder block → verify composite linking works (If→Else→End)
4. Add an endless block → verify same
5. Verify display names are correct


Greg-Orca:
Tested on VSN1 and TEK2
Action blocks and code works.

PR #1505, opened 2026-06-02
Reactivity performance improved for user_input and heartbeat handler

SukuWc:


narayb:
Feels similar to the old version, even on slower computers. During testing I haven't noticed problematic slowdowns.

PR #1537, opened 2026-06-23
More agile code block edit flow (#1390)

### LINKED ISSUE #967: Ctrl+S should save total state of Editor/Modules?
Ctrl+S should do something that creates a safety net.

### LINKED ISSUE #1390: Feature: More agile Code block edit flow
Closing and reopening the embedded code editor has been the biggest time sink I've encountered working with code in Grid Editor. 

When you want to open a code block in the editor and the editor already has another code block opened you need to close that one first. There's two improvements that would make working with code feel much nicer:

1. If the code that is already opened doesn't need "committing", it should be automatically closed and replaced by a new code block when you double-click it.
2. Making the code preview window editable without having to necessarily open the full view. This has the plus of helping you switch between to blocks easily - editing back and forth many times between two blocks is painful today.

SukuWc:
Closes #1390
Closes #967

Makes working with code blocks less of a time sink by removing the need to open/close the full editor modal to switch between blocks.

## Part 1 — auto-close & replace
A new `monaco-session.store` tracks the open editor's action + dirty state. Opening another block's editor while the current one is **clean** auto-closes and replaces it; while it has **unsaved changes** it's protected and the trigger is disabled.

## Part 2 — inline editing
The read-only preview is replaced by a real inline Monaco editor in the action block, so you can edit any number of blocks without ever opening the modal.

- `Monaco.svelte` decomposed into reusable **`CodeEditor`** (editor + validation + commit), **`CommitStatus`**, and **`CharacterCount`** components, reused by both the modal and the inline editor.
- Inline editor: **Commit**/**Discard** with the modal's commit behavior, auto-sizing (5–20 lines), syntax validation, no line numbers, horizontal scroll, suggestions off.
- **Open Editor** carries uncommitted inline edits into the modal; **discard** from the modal reverts the inline editor; **commit** from either keeps the other in sync.
- The block open in the modal is locked inline with a dimmed grayscale overlay.
- Syntax errors are dispatched to `DynamicWrapper` so the block is flagged invalid (red border) like other blocks, and cleared on teardown so the flag doesn't outlive the editor.
- "Powered by Lua" footer moved into the modal.
- Replaced `svelte/animate` `flip` with a translate-only `reflow` animation so blocks whose height changes (e.g. showing a syntax error) aren't vertically squished mid-animation.
- Border/background switched to grid-uikit theme tokens (`background-muted` / `background-soft`), also applied to `LineEditor`.

## Ctrl+S to commit/save (#967)
Ctrl/Cmd+S now commits/saves in every code-editor context. Each container catches the keystroke (`preventDefault` + `stopPropagation`, so it's scoped and doesn't hit the browser default or other handlers) and triggers the context's canonical button, which no-ops on its own when disabled:

| Context | Component | Action |
|---|---|---|
| Inline code block | `CodeBlock.svelte` | Commit |
| Code editor modal | `Monaco.svelte` | Commit |
| File Manager | `FileManager.svelte` | Save |

Per the direction in #967, this is scoped to editor/file-system commits rather than a global "save everything" state. Buttons are triggered via their rendered `<button>` for now; follow-up intechstudio/grid-uikit#50 proposes a `MoltenPushButton.trigger()` method to drop the wrapper + `querySelector`.

## Notes
- The literal "double-click the preview to open" trigger from #1390 is replaced by inline editing + an **Open Editor** button (the preview no longer exists).
- Uncommitted inline edits live only in the editor buffer until Commit; collapsing/navigating away discards them (visible "Unsaved changes!" status + Commit/Discard mitigate this).

### DIFF
--- a/playwright-tests/data/actionBlockElements.json
+++ b/playwright-tests/data/actionBlockElements.json
@@ -65,7 +65,7 @@
     "Encoder Push & Rotate": ["push rotate", "just rotate", "end"]
   },
   "code": {
-    "Code Block": ["input", "Edit Code"],
+    "Code Block": ["input", "Open Editor"],
     "Comment Block": ["input"],
     "Element Name": ["input"]
   },
--- a/playwright-tests/data/actionBlockLocators.js
+++ b/playwright-tests/data/actionBlockLocators.js
@@ -296,8 +296,8 @@ export const blocks = (page) => ({
     "Code Block": {
       block: page.locator("#action-menu").getByText("Code Block"),
       elements: {
-        input: page.locator("pre"),
-        "Edit Code": page.getByRole("button", { name: "Edit Code" }),
+        input: page.locator("code-block .view-line").first(),
+        "Open Editor": page.getByRole("button", { name: "Open Editor" }),
       },
     },
     "Comment Block": {
--- a/playwright-tests/pages/configPage.js
+++ b/playwright-tests/pages/configPage.js
@@ -189,7 +189,10 @@ export class ConfigPage {
   // Element and Action Operations
 
   async selectAllActions() {
-    await this.configFocus.click();
+    // Focus the config panel itself rather than clicking into the area (which
+    // can land in an inline code editor that captures Ctrl+A as select-text),
+    // so the select-all-actions shortcut actually fires.
+    await this.page.locator(".configpanel").first().focus();
     await this.keyboardActions.selectAll();
   }
 
--- a/playwright-tests/pages/configPage.js
+++ b/playwright-tests/pages/configPage.js
@@ -76,9 +76,18 @@ export class ConfigPage {
       .filter({ hasText: "End" })
       .locator("action-placeholder div")
       .first();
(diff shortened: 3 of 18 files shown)


Greg-Orca:
shrink the error messages box

Screenshot 2026-06-24 at 14 06 03

SukuWc:
@Greg-Orca fixed

Greg-Orca:
Screenshot 2026-06-29 at 11 22 47

This error occurred during testing, but I can no longer reproduce it. I used virtual module, I created two Action Blocks on a single element and introduced syntax errors in both. I opened the second one, corrected it, and committed the changes. After that, these error messages occurred.

Greg-Orca:
Unnecessary messages appear in the console
Screenshot 2026-06-29 at 11 34 17

SukuWc:
Automatic height calculator removed from inline code editor, height is now fixed 128px. The automatic height calculation is the most likely cause of the resize observer infinite loop. (Cannot verify 100%, let's hope)

PR #1538, opened 2026-06-23
Add startup window state preference (#1367)

### LINKED ISSUE #1367: Start editor Minimized
I'd like to start Grid Editor on startup since it's an integral part of my setup due to packages that require it to be open.

Either through the preferences or an executable argument I can add to my startup script, or better yet both!

SukuWc:
Closes #1367.

## What

Adds a **Startup window state** radio in **Preferences → General** that controls how the window appears when Grid Editor launches:

| Option | Behavior |
|---|---|
| **Open in a normal window** | Show normally (default) |
| **Open minimized to the taskbar** | `minimize()` — sits in the taskbar/dock |
| **Open hidden in the tray** | Stays hidden; tray icon on Windows/Linux, dock icon on macOS |

This addresses the request to start the editor minimized/unobtrusively (useful when it auto-starts with the system). The "if already running, open the window" half was already handled by the single-instance lock and is hardened here.

## How

- New persistent setting `startupWindowState` (`"normal" | "taskbar" | "tray"`, default `"normal"`) in `app-helper.store.ts`, persisted via the existing app-settings machinery.
- `Preferences.svelte`: a `MeltRadio` matching the adjacent "Run application in background" control.
- `main.ts`:
  - Window is created with `show: false` and shown on `ready-to-show` (also removes the white flash on normal launch); the chosen startup state is applied there.
  - `second-instance` handler now **restores + shows + focuses** the window on all platforms (previously `show()` was skipped on macOS, and only *minimized* windows were restored), so a minimized/tray start is always recoverable by relaunching. The deeplink parse is guarded so a plain relaunch no longer throws on `new URL()`.

## Testing

- `npm run web-build` passes.
- Verified on Linux: normal launch shows the window; relaunching a running instance triggers the single-instance handler without errors and re-focuses the existing window.
- ⚠️ **Needs cross-platform confirmation:** the `taskbar` (minimize) and `tray` (hide) paths could not be visually verified on the dev machine's tiling window manager (no tray/taskbar/minimize support there). Please confirm on **Windows** and **macOS** that those two options land the window minimized/hidden as intended and that it restores correctly.


Greg-Orca:
Mac OS ok

narayb:
Tested on Windows 11:

- Minimized to taskbar works as expected.
- Minimized to tray doesn't work, the application doesn't appear on the tray at all, nor does it start normally. _It potentially requires the keep app minimized after closeing the window setting on ON?_

Tested on Fedora Workstation 44 (GNOME, Wayland):

- Minimized to taskbar doesn't work, because there is no taskbar, AFAIK no Wayland window is created. Because the app is running, but no window was created, there's no way to change the option back in the GUI, so the user has to fix this in the `config.json`.
- Minimized to tray also doesn't work for the same reasons, no tray exists on GNOME DE. Because of this no Wayland window is created and the GUI app stays inaccessible. Once again, has to be fixed in the `config.json`.

PR #1548, opened 2026-07-02
Enhance: add flatpak linux target

kkerti:


narayb:
Consider removing the tar.gz from the flatpak archive to make the flatpak distribution smaller for nightly builds.

kkerti:
Works internally among machines. Further work will happen, once flatpak is released once. Then we can add it to flathub.

Issue #1536, opened 2026-06-22
bug: Element rename adds an empty action block


Greg-Orca:
**Describe the bug**
Element rename adds an empty action block to other element

<img width="407" height="295" alt="Image" src="https://github.com/user-attachments/assets/f568d6f3-0ecd-417c-86e1-593c7aa424a9" />

**To Reproduce**

1. Go to element 1 and rename it
2. Click on element 2
3. Element name 2 setup> empty Element Name action block appears

PR #1549, opened 2026-07-09
Add system MIDI channel selection and handling logic

Greg-Orca:
System element new dropdown. its adds the function action block and return the selected channel.
if auto then delete the function block

some bug: 
1. if close to the script length limit and I want to select a channel, error occurs 
2. if close to the script length limit, editing become really slow.


SukuWc:
closes #1536

Greg-Orca:
Find Element Name all position.

PR #1562, opened 2026-07-16
Block element/event navigation while CodeBlock has uncommitted changes

SukuWc:
- Add code-block-dirty.store with Set-based tracking for multiple instances
- Guard UserInput.set/update/flushPending with ConfirmModal prompt
- Revert radio/dropdown selection when navigation is blocked
- Block collapsing DynamicWrapper when code block has unsaved edits


SukuWc:
Contect menu/Reset to Defaults does not work when CodeBlock has uncommitted changes

SukuWc:
https://github.com/intechstudio/grid-editor/issues/1564

PR #1567, opened 2026-07-23
Landing update

Greg-Orca:


src/renderer/config-blocks/icons/element-name.svg
kkerti, line 1: Broken path, may look wrong. Fix path.

src/renderer/config-blocks/icons/end.svg
kkerti, line 1: This is a loop. Why is this the end?

src/renderer/config-blocks/icons/global.svg
kkerti, line 1: Hint: consider separating our own icon pack vs existing common icons: https://lucide.dev/icons/globe

May be a future task.

src/renderer/config-blocks/icons/lookup.svg
kkerti, line 1: Why not something with a https://lucide.dev/icons/search ? we take X value "look it up", then exchange it...

Yea exchange icon works as well, but then make sure "replace" or other common functionality does not collide.

src/renderer/config-blocks/icons/repeater-loop.svg
kkerti, line 1: Hah, this is "end" as well.

src/renderer/config-blocks/icons/simple-color.svg
kkerti, line 1: Get better tracing for this icon, bottom is jagged.

src/renderer/config-blocks/_composite-factory.ts
```diff
@@ -62,6 +62,7 @@ function buildPart(
     // Optional fields.
     menuName: data.menuName,
     helperText: data.helper,
+    description: data.description,
```
kkerti, line 65: Where is `helperText` applied, is `description` similar? These descriptions _do help_ understanding specific features, but are not strictly required for the context of software usage.

Nit pick, but maybe this would be better named as `helperDescription`?

src/renderer/config-blocks/_composite-groups.ts
```diff
@@ -86,6 +90,7 @@ export const compositeGroups: CompositeGroup[] = [
         menuName: "Encoder Push & Rotate L R",
         helper:
           "Actions here are triggered by rotating the encoder left while it is pressed.",
+        description: "Actions here are triggered by rotating the encoder left while it is pressed.",
```
kkerti: Nit pick: are we `rotating` or `turning` the encoder? The "turn" may be more user friendly.

If agreed, apply to all text.

src/renderer/config-blocks/ButtonStep_End.svelte
```diff
@@ -17,7 +18,7 @@
     defaultLua: "end",
     icon: undefined,
     blockIcon: undefined,
-    color: "#4A4AA7",
+    color: categoryColors["special"] as any,
```
kkerti: `as string`? why is `as any` needed?

When I see something like this, I am afraid some hacky thing is going on.

src/renderer/config-blocks/ButtonStep_If.svelte
```diff
@@ -32,7 +35,7 @@
     ],
     icon: icon,
     blockIcon: icon,
-    color: "#4A4AA7 ",
+    color: categoryColors["special"] as any,
```
kkerti: Same "as any", check the interfaces.

src/renderer/config-blocks/categoryColors.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Maps action block categories to their CSS variable.
+ * Colors are defined in app.css as --category-* variables,
+ * so you can retheme all blocks in one place.
+ */
+export const categoryColors: Record<string, string> = {
```
kkerti, line 6: This follows grid-uikit pattern for color variables? Do we use such thing in Editor elsewhere?

src/renderer/config-blocks/CodeBlock.svelte
```diff
@@ -10,22 +11,12 @@
     rendering: "standard",
     category: "code",
     displayName: "Code Block",
-    color: "#887880",
+    description: "Write Lua code by hand",
```
kkerti, line 14: These texts are not the description of the action block itself, instead they are helper informations. Isn't this field then redundant and we could get away by using `helperText`?

src/renderer/config-blocks/Midi.svelte
```diff
@@ -513,7 +495,5 @@
     </BlockRow>
   </Block>
 
-  <div class="mt-2">
-    <SendFeedback feedback_context="Midi" />
-  </div>
+  <div class="mt-2"></div>
```
kkerti, line 498: I agree that sendfeedback was underused, and contributed to visual clutter, but its worth a short memo as the _core idea_ is good. 

We should encourage people to share feedback and report bugs. The place for this is our public github, support mail, discord and the forum. Among these the github issue is most respected.

Anyway we still want to hear from people, consider adding a bug report memo to the welcome splash screen.

src/renderer/main/components/FirmwareDownloadOption.svelte
```diff
@@ -21,7 +21,9 @@
 </script>
 
 <BlockRow>
-  <BlockTitle>{title}</BlockTitle>
+  <div style="color: var(--foreground-muted);">
```
kkerti, line 24: So why is here a special coloring code instead of tailwind classes?

src/renderer/main/grid-layout/grid-modules/underlays/ElementSelection.svelte
```diff
@@ -127,6 +129,22 @@
       color-mix(in srgb, var(--foreground) 10%, var(--background));
   }
 
+  /* LCD content sits beneath this interaction layer, so use an outline instead
```
kkerti, line 132: Why does LCD content sit beneath the interaction layer? 

This feels like a workaround.

I don't like element specific css here, if we were able to handle element selection for **all** other elements, look into the DOM model how LCD element is added and check if it's possible to change it, so that the common css here applies to LCD as well.

src/renderer/main/grid-layout/grid-modules/underlays/ModuleBorder.svelte
```diff
@@ -39,6 +40,9 @@
     $appSettings.persistent.disableAnimations,
     $reduced_motion_store,
   );
+
+  $: showFirmwareMismatch =
+    !!device?.fwMismatch && $appSettings.firmwareNotificationState !== 0;
```
kkerti: I mean I get that it's cool to write one liners, but what does `!!` mean? 

`Boolean(device.fwMismatch)` might be more readable. 

Also why do we need to make sure fwMismatch is actually boolean? Can that valeue be  null or undefined as well and need this type enforcement?

src/renderer/main/grid-layout/grid-modules/Device.svelte
```diff
@@ -537,7 +537,7 @@
 
   :root {
     --element-margin: 5px;
-    --grid-rounding: 5px;
+    --grid-rounding: var(--radius);
```
kkerti, line 537: I did not see the overall anti-rounded patch yet, but I think this should be checked with @narayb!

src/renderer/main/modals/Welcome.svelte
kkerti, line 1: Appreciate the changes here, BUT there is work in relation to youtube fallback errors on this branch:

https://github.com/intechstudio/grid-editor/pull/1561

The above branch should arrive to stable first and be forward integrated here, as it is a smaller piece of work.

src/renderer/main/panels/configuration/components/DynamicWrapper.svelte
```diff
@@ -331,13 +335,42 @@
...
               </div>
 
-              <button
-                slot="edit-name-trigger"
-                on:click|stopPropagation={handleEditClicked}
-                class="cursor-pointer hover:bg-black/25 flex w-fit h-fit p-1.5 rounded pointer-events-auto"
-              >
-                <SvgIcon iconPath="edit" fill="#FFF" width={13} height={13} />
-              </button>
+              <div slot="edit-name-trigger" class="flex items-center gap-1">
```
kkerti, line 338: If you don't want to render dom element into a `slot`, consider using `svelte:fragment`.

Might be irrelevant here, but check it out: https://svelte.dev/docs/svelte/legacy-svelte-fragment

src/renderer/main/panels/configuration/components/Options.svelte
```diff
@@ -23,14 +23,15 @@
...
-    ? 'border-error'
-    : 'border-foreground'}  flex w-6 h-6 items-center justify-center border transition-opacity rounded-md fill-foreground"
+    ? 'border-foreground-disabled'
+    : 'border-foreground'}  flex w-6 h-6 items-center justify-center border transition-opacity fill-foreground"
   class:border-opacity-80={selected}
   class:bg-background-muted={selected}
   class:border-opacity-30={!selected}
   class:hover:border-opacity-100={!selected && !disabled}
-  class:bg-error={disabled}
+  style:background={disabled ? "var(--foreground-disabled)" : undefined}
```
kkerti, line 34: Yea, I would probably rework our tailwind theme and config than mix n match styling this way.

src/renderer/main/panels/configuration/Configuration.svelte
kkerti, line 1: This has now two great additional component candidates:
- ElementNameChangePanel.svelte
- ModuleMIDIChannelChangePanel.svelte

Organizing these two into components would make the Configuration.svelte much leaner. The interfaces of those components would become clearer too.

Review the comments for both module midi channel change and element name work, I agree that some logic should be commented, but ideally the code should document itself.

kkerti: Also those file names are first guesses, they can be differently!

src/renderer/main/panels/configuration/ElementSelectionPanel.svelte
```diff
@@ -1,14 +1,18 @@
...
   import { get } from "svelte/store";
-  import { MeltSelect } from "@intechstudio/grid-uikit";
+  import { MeltSelect, MoltenInput } from "@intechstudio/grid-uikit";
   import { user_input } from "./../../../runtime/user-input.store";
   import { appSettings } from "./../../../runtime/app-helper.store";
   import { GridPage, PageData } from "../../../runtime/runtime";
+  import { tick } from "svelte";
 
   export let page: GridPage;
+  export let isEditingName: boolean = false;
```
kkerti, line 10: Ok, configuration has `isEditingElmentName`, something seems redundant / duplicated. Review how these panels should be interefaced with one and other (if any interfacing is even needed).

src/renderer/runtime/app-helper.store.ts
```diff
@@ -195,6 +196,14 @@ async function init_appsettings() {
             value = persistentDefaultValues[key];
           }
 
+          if (
```
kkerti, line 200: interesting, we'll talk about themes!

src/renderer/app.css
```diff
@@ -3,6 +3,23 @@
 @tailwind components;
 @tailwind utilities;
 
+/* Stroke-only icons: prevent fill from being inherited from parent containers */
```
kkerti, line 6: Why?

kkerti (commented):
Great work, with only tiny changes Editor feels better. 

Couple highlights to discuss beside my comments:
- theming with native css vs. tailwindcss
- component creation for Configuration.svelte
- description vs helper

src/renderer/config-blocks/_composite-groups.ts
Greg-Orca: following the menuname: Encoder Push & Rotate L R

src/renderer/main/panels/configuration/Configuration.svelte
Greg-Orca: created the two component ElementNameChangePanel.svelte
ModuleMIDIChannelChangePanel.svelte

src/renderer/app.css
Greg-Orca: helps to keep transparent

src/renderer/config-blocks/headers/MidiSysExFace.svelte
Greg-Orca, line 1: align info box right

src/renderer/config-blocks/icons/end.svg
Greg-Orca, line 1: delete icon and where we use it

src/renderer/main/user-interface/ActiveChanges.svelte
```diff
@@ -255,13 +256,34 @@
       });
     });
   }
+
+  function handleWindowKeyDown(e: KeyboardEvent) {
```
Greg-Orca: this should be defined in app.svelte

PR #1580, opened 2026-08-06
fix(file-manager): Lua file upload — shortify + minify on save, expand on read

kkerti:
The file manager tagged `.lua` files as `intech_lua`, but the save/read transforms compared `selectedLanguage === "lua"` — never true. So human-readable Grid API names (e.g. `draw_rectangle_filled`) were uploaded to modules verbatim and failed at runtime (`attempt to call a nil value`), and device files were shown without expanding.

**Fix:** route `.lua` save through `GridScript.compressScript` (shortify + minify — compact storage for memory-constrained flash) and read through `GridScript.expandScript` (humanize + beautify) — the same pipeline the runtime uses to send code to a module. The language id is centralized in one constant so the language map and the transform gates can't drift apart again.

### DIFF
--- a/src/renderer/main/panels/FileManager/FileManager.svelte
+++ b/src/renderer/main/panels/FileManager/FileManager.svelte
@@ -26,7 +26,7 @@
     closeEditorContext,
   } from "../../../lib/monaco-luals-client";
 
-  // Monaco language id for Grid Lua files. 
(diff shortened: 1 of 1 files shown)


kkerti:
TEST: copying human readable function names with sequencer project works now, we should still do focused test with @Greg-Orca

PR #1603, opened 2026-08-28
Fix restart on Flatpak

kkerti:
On Flatpak the app is started through `zypak-wrapper`, which sets up Chromium's sandbox inside the container. A bare `app.relaunch()` re-executes `process.execPath` directly and skips that wrapper, so the relaunched instance fails sandbox init and the app never comes back. Relaunch through `/app/bin/zypak-wrapper` instead.

### DIFF
--- a/src/electron/main.ts
+++ b/src/electron/main.ts
@@ -22,16 +22,21 @@ import chokidar from "chokidar";
 import configuration from "../../configuration.json";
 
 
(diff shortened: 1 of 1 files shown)


kkerti:
Moved to the flathub work branch (enhance/add-snap-and-flatpak-linux-targets).

